PR #1, opened 2021-02-17
Fix meld logic when handling oneof and optional values.

kku1993:
Fixes:
- Collapse optional and non-optional values if identical 
    ````
    meld(optional<T>, T) => optional<T>
    ````

- Make all oneofs additive, not just for Akita injected conflicts
    ````
    meld(oneof<T1, T2>, oneof<T1, T3>) => oneof<T1, T2, T3>
    meld(oneof<T1, T2>, T3) => oneof<T1, T2, T3>
    ````

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -122,11 +122,10 @@ func MeldData(dst, src *pb.Data) (retErr error) {
 		}
 	}()
 
-	// Check if src is already a oneof that encodes a conflict. This can happen if
-	// src is the collapsed element from a list originally containing elements
-	// with conflicting types.
-	if srcOf, ok := src.Value.(*pb.Data_Oneof); ok && srcOf.Oneof.GetPotentialConflict() {
-		if v, ok := dst.Value.(*pb.Data_Oneof); ok && v.Oneof.GetPotentialConflict() {
+	// Check if src is already a oneof. This can happen if src is the collapsed
+	// element from a list originally containing elements with conflicting types.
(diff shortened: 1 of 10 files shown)


thatplguy:
Quick question before I run into another meeting: Does this handle the commutative case of the operations described above, e.g.

* `meld(T, optional<T>) => optional<T>`
* `meld(T3, oneof<T1, T2>) => oneof<T1, T2, T3>`

Also, is it possible to have recursive `oneof`?  E.g. `meld(optional<T>, optional<optional<T>>)`?

kku1993:
> Quick question before I run into another meeting: Does this handle the commutative case of the operations described above, e.g.
> 
> * `meld(T, optional<T>) => optional<T>`
> * `meld(T3, oneof<T1, T2>) => oneof<T1, T2, T3>`
> 

Yes, the unit tests run the list of inputs in both forward and reverse directions to test this. [1]

> Also, is it possible to have recursive `oneof`? E.g. `meld(optional<T>, optional<optional<T>>)`?

I think you mean recursive `optional`, not `oneof`. In practice, we should never get `optional<optional<T>>` because:
1. Witness generation only creates `optional<T>`
2. The logic in meld (this PR) ensures that we don't create recursive `optional`

[1] https://github.com/akitasoftware/akita-libs/blob/32bfc2802d768ccfb461a83ef7960bfb4ed58f63/spec_util/meld_test.go#L95

thatplguy (approved):

PR #2, opened 2021-02-18
Fix meld logic for optional non-primitives.

kku1993:
In the scenario `meld(T, optional<T>)`, the change in #1 only handled
the case where `T` is a primitive. This PR fixes the logic to handle
complex types such as `struct` and `lists`.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -110,6 +110,18 @@ func mergeExampleValues(dst, src *pb.Data) {
 	}
 }
 
+func makeOptional(d *pb.Data) {
+	if !isOptional(d) {
+		d.Value = &pb.Data_Optional{
+			Optional: &pb.Optional{
+				Value: &pb.Optional_Data{
+					Data: &pb.Data{Value: d.Value},
+				},
+			},
+		}
+	}
+}
+
 // Assumes that dst.Meta == src.Meta.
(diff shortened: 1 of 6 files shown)


spec_util/meld.go
```diff
@@ -184,13 +187,8 @@ func MeldData(dst, src *pb.Data) (retErr error) {
...
 		case *pb.Optional_Data:
-			// Check whether src == dst with optional removed.
-			// E.g. meld(optional<string>, string) => optional<string>
-			if dataEqual(opt.Data, src) {
-				// Do nothing - src is just non-optional verison of dst.
-				return nil
-			}
-			return recordConflict(dst, src)
+			// Meld src with the non-optional version of dst.
+			return MeldData(opt.Data, src)
```
thatplguy, line 191: Do you have to call `makeOptional` on the result here, or otherwise the optional will be discarded?

thatplguy (approved):

spec_util/meld.go
kku1993: No, because we are calling `MeldData` on the thing *inside* of the optional here, so the outer optional is retained.

thatplguy: Ah, right, because this overwrites `dst` rather than producing a new data object.  Got it.

PR #3, opened 2021-02-20
Include spec name in metadata

thatplguy:
Adds the spec name to the GetSpecResponse metadata, and adds a new response
type GetSpecMetadataResponse, which fetches the spec metadata without including
the spec itself.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -132,6 +132,23 @@ type CreateSpecResponse struct {
 	ID akid.APISpecID `json:"id"`
 }
 
+type GetSpecMetadataResponse struct {
+	// TODO: remove
+	// If the spec was created from a learn session, the session's ID is included.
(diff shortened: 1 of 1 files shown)


api_schema/schema.go
```diff
@@ -144,6 +161,8 @@ type GetSpecResponse struct {
 	// IDs are included.
 	LearnSessionIDs []akid.LearnSessionID `json:"learn_session_ids,omitempty"`
 
+    Name string `json:"name"`
```
kku1993, line 164: nit: indentation is off, make sure your editor uses actual tabs rather than spaces (it's a go thing)

kku1993 (approved):

api_schema/schema.go
thatplguy: Thanks, fixed.

PR #6, opened 2021-02-25
Ensure examples are stable

thatplguy:
This fixes a bug where examples chosen for path parameters depended on the
order in which witnesses were processed.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -106,36 +106,31 @@ func isNone(d *pb.Data) bool {
 func mergeExampleValues(dst, src *pb.Data) {
 	examples := make(map[string]*pb.ExampleValue, 2)
 
-	// Get up to two examples each from src and dst.
+	// Get all (unique) example keys.
+	keySet := make(map[string]struct{}, len(src.ExampleValues)+len(dst.ExampleValues))
 	exampleMaps := []map[string]*pb.ExampleValue{dst.ExampleValues, src.ExampleValues}
 	for _, exampleMap := range exampleMaps {
-		keys := make([]string, 0, len(exampleMap))
 		for k := range exampleMap {
-			keys = append(keys, k)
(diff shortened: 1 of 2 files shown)


spec_util/meld.go
```diff
@@ -117,11 +117,25 @@ func mergeExampleValues(dst, src *pb.Data) {
 		// Sort keys to ensure a stable selection.
 		sort.Strings(keys)
```
kku1993, line 118: Instead of doing 2 loops, how about we just collect all the keys from src and dst into 2 arrays, sort both, then take the first element from each?

kku1993 (changes requested):

spec_util/meld.go
thatplguy: That's basically what the old code did.  The problem is that if we're merging, say, `["ex1", "ex2"]` and `["ex3", "ex4"]`, we want `["ex1", "ex2"]`, not `["ex1", "ex3"]`.

I did it this way to avoid looping over all the elements of each example map.  On the other hand, it should be small unless a user uploads a spec with a lot of examples, so a simpler approach might be to add all examples from `src` and `dst` to `examples`, sort the keys, and remove all but the smallest two.  Thoughts?

kku1993: > I did it this way to avoid looping over all the elements of each example map. On the other hand, it should be small unless a user uploads a spec with a lot of examples, so a simpler approach might be to add all examples from src and dst to examples, sort the keys, and remove all but the smallest two. Thoughts?

Yeah let's do this. I don't think you even need to add all examples together, just get all the keys, get the smallest 2, and find them in either `src` or `dst`

thatplguy: Sounds good to me.  Done.

kku1993 (approved):

PR #13, opened 2021-04-08
Update martian version

mgritter:
This is the same change as https://github.com/akitasoftware/akita-cli/pull/10; it feels wrong to have to make it both places? 

Maybe cutting a release on martian would be the way to go.


thatplguy (approved):
Something to consider for the future.  I don't think we'll be making changes that often though, so I'm fine with a little inefficiency when we do, at least for now.

PR #17, opened 2021-04-15
Allow object names to be omitted from Akita URIs

liujed:
This allows us to say `akita://service:spec` and `akita://service:trace` for cases where we want object names to be auto-generated.

### DIFF
--- a/akiuri/akiuri.go
+++ b/akiuri/akiuri.go
@@ -45,7 +45,11 @@ type URI struct {
 }
 
 func (u URI) String() string {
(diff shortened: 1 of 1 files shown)


liujed:
Looks like CircleCI is failing because it's not configured for this project.

thatplguy (approved):

mgritter (approved):
Looks good.  I did see a successful CircleCI run on main, but now it's back to complaining that the config doesn't exist.

PR #18, opened 2021-04-16
Add support for "akita://service" URIs

liujed:
This changes the representation of Akita URIs: the `ObjectType` component is now a pointer and is `nil` when the object type is not given. Also removed the `UNKNOWN_TYPE` object type, which should not have been externally visible.

### DIFF
--- a/akiuri/akiuri.go
+++ b/akiuri/akiuri.go
@@ -1,5 +1,12 @@
 package akiuri
 
+// There are three kinds of URIs supported here:
+//  * akita://serviceName
+//      For these, the ObjectType will be nil and the ObjectName will be empty.
+//  * akita://serviceName:objectType
+//      For these, the ObjectName will be empty.
+//  * akita://serviceName:objectType:objectName
+
 import (
 	"fmt"
 	"strings"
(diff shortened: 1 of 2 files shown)


akiuri/akiuri.go
```diff
@@ -12,19 +19,26 @@ const (
...
-	SPEC
-	TRACE // aka learn session
+	SPEC  ObjectType = iota
+	TRACE            // aka learn session
 )
 
-func stringToObjectType(s string) ObjectType {
+func (o ObjectType) Ptr() *ObjectType {
+	return &o
+}
```
mgritter, line 28: I like everything about this but .Ptr() but I can't come up with an alternative I like more, so I guess we're going with it.

What I was thinking about: 

1. export ObjectType as a pointer type instead of int? I don't think we can have consts that are addresses.

2. export an interface type

mgritter (approved):

akiuri/akiuri.go
```diff
@@ -12,19 +19,26 @@ const (
...
+
+func (o1 *ObjectType) Is(o2 ObjectType) bool {
+	return o1 != nil && *o1 == o2
+}
```
mgritter, line 32: Another alternative is to have IsTrace() and IsSpec() methods, though Go stdlib has been moving more towards a model like this for I/O errors.  But right now your code in the other PR looks like

Is(akiuri.Trace)

instead of

IsTrace()

akiuri/akiuri.go
thatplguy: Perhaps an explicit `NONE` object type, rather than using `nil` to represent it?  Then there's no need to use pointers at all.

That being said, I think this is one of those things where this implementation gets the job done 🙂 .  I'm happy with it going in as is.

thatplguy (approved):

akiuri/akiuri.go
liujed: I thought about that, but then the implementation of `String()` becomes janky. Not clear what to return there. The current implementation side-steps the issue by requiring an `ObjectType` instance before you can call `String()`.

liujed: I also considered using option types in the client code, but apparently, those aren't really a thing in Go.

PR #21, opened 2021-04-28
Add utility for computing spec hash keys

thatplguy:
Some IR objects (Method.Args, Method.Responses, and Data.OneOf) require that
their map keys be hashes of the corresponding data values, e.g. map[hash(v)] =
v.  This PR adds a utility function for recomputing hash keys.

It also updates some unit test files that, it turns out, did not respect this
invariant.

### DIFF
--- a/spec_util/hash_keys.go
+++ b/spec_util/hash_keys.go
@@ -19,8 +19,16 @@ func (vis *hashOneOfVisitor) VisitData(c http_rest.HttpRestSpecVisitorContext, p
 	if oneOf == nil {
 		return true
 	}
+
 	options := p.GetOneof().Options
-	for k, v := range options {
+	keys := make([]string, 0, len(options))
+
+	for k, _ := range options {
+		keys = append(keys, k)
+	}
+
+	for _, k := range keys {
+		v := options[k]
 		h, err := pbhash.HashProto(v)
 		if err != nil {
 			vis.err = err
@@ -52,7 +60,12 @@ func RewriteHashKeys(spec *pb.APISpec) error {
 	// Hash Args and Responses for each method.
 	for _, method := range spec.Methods {
 		for _, m := range []map[string]*pb.Data{method.Args, method.Responses} {
-			for k, arg := range m {
+			keys := make([]string, 0, len(m))
+			for k, _ := range m {
+				keys = append(keys, k)
+			}
+			for _, k := range keys {
+				arg := m[k]
 				h, err := pbhash.HashProto(arg)
 				if err != nil {
 					return errors.Wrap(err, "failed to compute hash of method argument")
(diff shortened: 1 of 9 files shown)


spec_util/hash_keys.go
```diff
@@ -0,0 +1,69 @@
...
+	options := p.GetOneof().Options
+	for k, v := range options {
+		h, err := pbhash.HashProto(v)
+		if err != nil {
+			vis.err = err
+			return false
+		}
+		if k != h {
+			delete(options, k)
+			options[h] = v
```
mgritter, line 31: It's not a big problem, but doing it this way means you may visit the same option twice (and hash it twice.)  Or not, it's not guaranteed.

This could be avoided by extracting the keys from the map first, and iterating over the original set of keys.

spec_util/hash_keys.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, method := range spec.Methods {
+		for _, m := range []map[string]*pb.Data{method.Args, method.Responses} {
+			for k, arg := range m {
+				h, err := pbhash.HashProto(arg)
+				if err != nil {
+					return errors.Wrap(err, "failed to compute hash of method argument")
+				}
+				if h != k {
+					delete(m, k)
+					m[h] = arg
```
mgritter, line 62: Same problem here, may get the same arg twice, once with k and once with h.

mgritter (approved):

spec_util/hash_keys.go
liujed: Indeed, mutating a data structure while you're iterating over it usually results in undesired behaviour.

spec_util/hash_keys_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/akitasoftware/akita-libs/test"
+)
+
+func TestHashKeyRewrite(t *testing.T) {
+	// We reuse test data from meld_test, because the expected witnesses
+	// cover a range of expected IR objects.
+	for _, testData := range tests {
+		fmt.Printf("COLE: %s\n", testData.name)
```
liujed, line 18: Did you intend to have this here?

liujed (approved):

spec_util/hash_keys_test.go
thatplguy: No I did not 😅.  Fixed in 70b9fba.

spec_util/hash_keys.go
thatplguy: Thanks, fixed in 70b9fba.

spec_util/hash_keys.go
thatplguy: Thanks, fixed in 70b9fba.

PR #24, opened 2021-05-10
Extended IR visitor framework

liujed:
This extends the visitor framework to provide more control over the traversal, and to add hooks for visiting metadata nodes. It also introduces a visitor for traversing a pair of nodes in tandem.

Pre- and post-order traversal has been generalized: the framework now exposes `enter` and `leave` methods for each node type. Preorder traversal can be obtained by implementing `enter` methods, whereas postorder traversal is obtained by implementing `leave` methods. The visiting of each node type's children can also be customized.

Visit methods now return an enum value rather than a `bool` to indicate how traversal should proceed. Possibilities are `Continue` to continue with normal traversal; `SkipChildren` allows the `enter` method to specify that the children of the current node should not be visited; `Stop` to stop the traversal, but execute the `leave` methods on the way back up to the root; and `Abort` halts the traversal immediately.

Added `enter`, `leave`, and `visitChildren` methods for metadata IR nodes.

### DIFF
--- a/visitors/go_ast_pair/ast_pair_visitor.go
+++ b/visitors/go_ast_pair/ast_pair_visitor.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"reflect"
 	"strconv"
+	"strings"
 	"unicode"
 
 	. "github.com/akitasoftware/akita-libs/visitors"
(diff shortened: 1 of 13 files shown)


visitors/go_ast_pair/ast_pair_visitor.go
```diff
@@ -0,0 +1,294 @@
...
+	"fmt"
+	"reflect"
+	"strconv"
+	"strings"
+	"unicode"
+
+	. "github.com/akitasoftware/akita-libs/visitors"
+)
+
+// Structurally recurses through `leftNode` and `rightNode` in tandem.
```
thatplguy, line 13: Is there a comment somewhere that describes what the pair visitor does when there are differences between the structures of the two nodes?

It looks like it descends until it finds a difference (either left or right are missing a node, or the nodes are of different types), and visits those nodes but not their children?

thatplguy (approved):

PR #25, opened 2021-05-13
Introduced diff visitors and improved extensibility of existing visitors

liujed:
A diff visitor is a pair visitor that provides hooks for handling diffs when they are detected between two IR trees.

Also improved extensibility of existing visitors and moved `dataComparator` to here from [akitasoftware/superstar](/akitasoftware/superstar).


visitors/http_rest_diff/data_comparator.go
```diff
@@ -0,0 +1,27 @@
...
+
+// A diff visitor for determining whether two Data instances are different.
+type dataComparator struct {
+	DefaultHttpRestSpecDiffVisitorImpl
+
+	isDifferent bool
+}
+
+func (v *dataComparator) EnterDiff(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}) Cont {
+	v.isDifferent = true
```
mgritter, line 25: I don't understand, why are two Data instances always different?

(Or, they're different only if they get to this case, DefaultHttpRestSpecDiffVisitorImpl covers everything else?)

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
```diff
@@ -0,0 +1,550 @@
...
+	LeaveChangedNode(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}, cont Cont) Cont
+}
+
+// An HttpRestSpecDiffVisitor implementation. This does not traverse into the
+// children of nodes that were added, removed, or changed.
+type DefaultHttpRestSpecDiffVisitorImpl struct {
+	http_rest.DefaultHttpRestSpecPairVisitor
+}
+
+var _ DefaultHttpRestSpecDiffVisitor = &DefaultHttpRestSpecDiffVisitorImpl{}
```
mgritter, line 79: A variant that also works is:

```
        var _ DefaultHttpRestSpecDiffVisitor = (DefaultHttpRestSpecDiffVisitorImpl *)(nil)
```

It looks like Vault uses both varieties.

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
```diff
@@ -0,0 +1,550 @@
...
+
+// An HttpRestSpecPairVisitor with hooks for processing each difference found
+// between two IR trees. A node is considered changed if a difference can be
+// observed at that level of the IR. For example, HTTPAuth nodes with different
+// Types are considered changed, but their parents might not necessarily be
+// considered changed.
+//
+// Go lacks virtual functions, so all functions here take the visitor itself as
+// an argument, and call functions on that instance.
+type HttpRestSpecDiffVisitor interface {
```
mgritter, line 23: Because http_rest_diff is already in the name of the package, I think it would be more idiomatic to use shorter names for the interfaces.  I guess it's too late to do that with http_rest but you can see further down how 

http_rest.SpecPairVisitorContext

conveys the same information as

http_rest.HttpRestSpecPairVisitorContext

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
```diff
@@ -0,0 +1,550 @@
...
+	HttpRestSpecDiffVisitor
+
+	EnterDiff(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}) Cont
+	LeaveDiff(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}, cont Cont) Cont
+
+	// Delegates to EnterDiff by default.
+	EnterAddedOrRemovedNode(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}) Cont
+
+	// Delegates to EnterDiff by default.
+	EnterChangedNode(self interface{}, ctx http_rest.HttpRestSpecPairVisitorContext, left, right interface{}) Cont
```
mgritter, line 64: So, I thought about this some more, and there is another option not yet discussed.  But since you're doing type assertions anyway it should be about equal cost (they are cheap in Go.)

Instead of creating a large interface, and then a base class to provide sensible defaults, you could create small interfaces. Then move that decision about what method to call elsewhere, so that instead of EnterAddedOrRemovedDiff having a "default" implementation, the visit method sees whether the visitor has the appropriate interface--- and if not, drop down to a simpler one. 

https://play.golang.org/p/FkRz48Zfjxe

I'm not 100% sure this is a fit, but it seems to solve this tactical problem of "how do I delegate within the interface" by moving that delegation to something that already has the original object.

You can even do this as an adaption layer, you could have one big interface made of lots of the small interfaces, and all the "what do I delegate to" logic lives within that.

mgritter (commented):
I will take a closer look a the nontrivial methods tomorrow.

visitors/http_rest_diff/data_comparator.go
liujed: This is only ever called on two `Data` instances that have a diff—either `left` was removed, `right` was added, or `left` was changed to `right`. (Elsewhere, `left` is called `baseNode` and `right` is called `newNode`. I'll fix these parameter names to be consistent.)

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
liujed: We already manually dispatch on the type of the node being visited: https://github.com/akitasoftware/akita-libs/blob/jed/diff-visitor/visitors/http_rest/http_rest_spec_pair_visitor.go#L436. I'm not convinced that additionally dispatching on the visitor type for each of those cases is going to be better.

I'm also inclined to leave things as they are, in the name of "scrappiness"; especially if the only thing gained from a refactoring is the badge of being idiomatic.

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
liujed: In 8147936.

visitors/http_rest_diff/http_rest_spec_diff_visitor.go
liujed: In 8147936.

mgritter (approved):

thatplguy (approved):
LGTM.

PR #27, opened 2021-05-15
Richer context representation & bug fixes

liujed:
Contexts now track the traversal path more richly. They have the set of nodes going back to the root, and a representation of the edge that was followed from each node.

Each context now has a pointer to its parent.

We can now obtain from the context the innermost `Data` node being visited, and that node's context.

Made `ExtendContext` completely imperative, rather than partially imperative and partially functional.

Just call `ExtendContext` once when visiting a node. Previously, we were calling it thrice: within `Enter`, before calling `VisitChildren`, and within `Leave`.

Fix: we weren't aborting traversal immediately if `VisitChildren` requested an abort.

Fix pair visitor: when visiting a node that only exists in one half of the pair, we were always calling `EnterDifferentTypes` and `LeaveDifferentTypes` because we were using a nil-typed `nil` for the missing value.

Gave `httpRestSpecVisitorContext` a shorter name.

### DIFF
--- a/visitors/go_ast/ast_visitor.go
+++ b/visitors/go_ast/ast_visitor.go
@@ -5,7 +5,6 @@ package go_ast
 import (
 	"fmt"
 	"reflect"
-	"strconv"
 	"strings"
 	"unicode"
 
@@ -32,6 +31,8 @@ func (t *astVisitor) visit(c Context, m interface{}) Cont {
 		return Continue
 	}
 
+	t.vm.ExtendContext(c, t.vm.Visitor(), m)
+
 	keepGoing := t.vm.EnterNode(c, t.vm.Visitor(), m)
 	switch keepGoing {
 	case Abort:
@@ -45,8 +46,17 @@ func (t *astVisitor) visit(c Context, m interface{}) Cont {
 
 	// Don't visit children if we are stopping or skipping children.
 	if keepGoing == Continue {
-		newContext := t.vm.ExtendContext(c, t.vm.Visitor(), m)
-		keepGoing = t.vm.VisitChildren(newContext, t.vm, m)
+		keepGoing = t.vm.VisitChildren(c, t.vm, m)
+		switch keepGoing {
+		case Abort:
+			return Abort
+		case Continue:
+		case SkipChildren:
+			panic("VisitChildren shouldn't return SkipChildren")
+		case Stop:
+		default:
+			panic(fmt.Sprintf("Unknown Cont value: %d", keepGoing))
+		}
 	}
 
 	keepGoing = t.vm.LeaveNode(c, t.vm.Visitor(), m, keepGoing)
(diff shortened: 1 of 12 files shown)


visitors/go_ast/ast_visitor.go
```diff
@@ -32,6 +31,8 @@ func (t *astVisitor) visit(c Context, m interface{}) Cont {
 		return Continue
 	}
 
+	t.vm.ExtendContext(c, t.vm.Visitor(), m)
```
thatplguy, line 34: If this mutates the context, do we need to undo the mutation before we return?

thatplguy (commented):
Looks good so far.  I'll finish reviewing in a bit -- I need to look back through how the context is used to make sure I understand the change.

visitors/go_ast/ast_visitor.go
mgritter: I have a similar question, from reading the PR it's not clear to me what the new discipline for using a Context is-- any information must be deep-copied out of it before returning?

thatplguy: Ah, I see.  When visiting children (with `go_ast.DefaultVisitChildren`), the context is copied by calls to `context.EnterStruct`, `context.EnterArray`, and `context.EnterMapValue`.  So when visiting a node `n`, the workflow is:

1. Mutate the context.
2. Call `EnterNode`.
3. Call `VisitChildren`, which copies the context and extends the copy before visiting each child.
4. Call `LeaveNode`.

This ensures that all the visitor methods see the updated context.  We don't need to undo the mutation because  either (a) the context is a copy and will be discarded before visiting any siblings, or (b) there are no siblings, in which case the context is no longer used.

It's equivalent to pushing/popping the context, except the push/pop happens immediately prior to visiting each child rather than on entry to a node.

thatplguy: @liujed is there an example somewhere of overriding one of the `Visit*Children` methods for an HTTP/REST visitor?  Is it possible to override one of those methods and forget to copy the context before visiting children?

thatplguy (approved):
Ok, I'm convinced this works as intended for our current use cases 🙂 .

There's some complexity under the hood (e.g. a lot of indirection, visitors calling visitor managers calling the context etc.) that results from Go not having virtual function pointers.  In the future, we can revisit this to make it easier to understand and to write new kinds of visitors.  I didn't do a great job making this clear in the original visitor infrastructure, and @liujed inherited my mistakes there.

For the time being, it's easy to use the visitor infrastructure (e.g. visit all primitives in a spec) but complicated to extend the infrastructure itself, which works since the former is the more common use.

visitors/go_ast/ast_visitor.go
liujed: Here's an example of overriding `Visit*Children` for the pair visitor. Currently, if you override `Visit*Children`, before visiting each child, you need to remember to create a context for that child. I suppose we could roll the two operations into `VisitStructField`, `VisitArrayElement`, and `VisitMapValue` methods that update the context for you.

thatplguy: What about having the visitor infrastructure copy the context before calling `Visit*Children`, and changing `context.EnterArray` etc. to mutate rather than copy?  That should result in roughly the same number of copy operations, while preventing users of the infrastructure from making a subtle mistake in forgetting to copy the context.

liujed: Not sure that's enough. Implementations of `Visit*Children` still need to make sure they visit each child with a fresh context, or else the context state will be tainted across sibling nodes.

PR #28, opened 2021-05-18
Added field path to `SpecVisitorContext`

liujed:
`SpecVisitorContext` now tracks the location of the innermost field being visited.

Also shortened a few filenames.


visitors/http_rest/field_location.go
```diff
@@ -0,0 +1,75 @@
+package http_rest
```
thatplguy, line 1: Can this go in `field_utils/field_location.go` (and update it to use the new path elements)?  It would be nice to have a single notion of FieldLocation.

https://github.com/akitasoftware/superstar/blob/c40faa6300b40ff0ea03721da290f20931d324d8/lib/field_utils/field_location.go

thatplguy: Or call this something like `type RelativeFieldLocation FieldLocationElement[]` and update the other `FieldLocation` struct to replace `path` with`relativeLocation: RelativeFieldLocation`?

visitors/http_rest/spec_visitor_context.go
```diff
@@ -25,6 +25,13 @@ func (t HttpValueType) String() string {
 type SpecVisitorContext interface {
 	visitors.Context
 
+	// Identifies the REST field being visited when combined with IsArg,
+	// IsResponse, and GetValueType. For fields outside of bodies and
+	// authorization headers, the first component of the path is a FieldName
+	// identifying the path argument, query variable, header, or cookie in which
+	// the field is located.
+	GetFieldLocation() []FieldLocationElement
```
thatplguy, line 33: `GetRelativeFieldLocation` to avoid confusion with the other definition of `FieldLocation`?

thatplguy (approved):
LGTM, aside from reconciling the two notions of field location.

visitors/http_rest/field_location.go
liujed: Ideally, we'd unify the two, but is this really the data structure we want to export to the front end?

liujed: To be clear, the `FieldLocationElement`s would be replacing the strings in this array: https://github.com/akitasoftware/superstar/blob/c40faa6300b40ff0ea03721da290f20931d324d8/lib/field_utils/field_location.go#L35-L40

visitors/http_rest/spec_visitor_context.go
liujed: Commit 940575f calls these "paths" instead of "locations".

visitors/http_rest/field_location.go
thatplguy: Yea, fair enough -- ideally we'd unify them and write a nice serializer, but it's not worth the effort right now.

PR #32, opened 2021-05-19
Melding structs with different keys should result in optionals

mgritter:
In the original test `favorite_animal` is present in one struct but not the other, so probably should be marked optional when melded.

I've repurposed this branch for the fix as well.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -449,11 +449,7 @@ func meldStruct(dst, src *pb.Struct) error {
 	// ... but what if we end up with hundreds of keys?
 	// Probably that means the key is part of the data. Let's use an arbitrary
 	// threshold, and if we cross it, merge all the keys.
-	// TODO: convert to a new Map type instead of a Struct? Very disruptive, have to fix all the visitors etc.
-	// or TODO: use a special key "*" (matching pq syntax?), but somebody could actually use that in practice
-	// or TODO: use a special key that is less likely to be used, but remap it when displaying it or showing
-	// it as an OpenAPI spec.
-	// or TODO: add another field to Struct specifying that this has occurred.
+	// TODO: add another field to Struct in in the IR specifying that this has occurred.
 	// (and if we can infer a data format on the keys, it could live at that level too.)
 
 	return nil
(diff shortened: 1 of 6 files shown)


thatplguy (approved):
This looks right to me.  As you pointed out, I'd expect a field present in some witnesses but not others to be marked as optional.

mgritter:
We could check this in as-is, but I'd like to fix the "100s of keys" case too.  I think the right thing is to add extra fields in the pb.Struct type, so we can distinguish between: (a) structs with fixed keys, the default, and (b) structs with an unbounded collection of keys, i.e., maps.

It looks like Swagger and OpenAPI makes this distinction, an object can have "properties" and "additionalProperties"

spec_util/meld.go
```diff
@@ -398,20 +418,44 @@ func assignDataFormats(d *pb.Data, formats map[string]bool) {
...
+
+	// ... but what if we end up with hundreds of keys?
+	// Probably that means the key is part of the data. Let's use an arbitrary
+	// threshold, and if we cross it, merge all the keys.
+	// TODO: convert to a new Map type instead of a Struct? Very disruptive, have to fix all the visitors etc.
+	// or TODO: use a special key "*" (matching pq syntax?), but somebody could actually use that in practice
+	// or TODO: use a special key that is less likely to be used, but remap it when displaying it or showing
+	// it as an OpenAPI spec.
+	// or TODO: add another field to Struct specifying that this has occurred.
+	// (and if we can infer a data format on the keys, it could live at that level too.)
```
thatplguy, line 457: I'm leaning toward that last option -- adding another field to Struct indicating that it's a data object, and optionally recording any data formats we can infer.  We could do both keys and values, potentially.  The struct itself could just be empty -- we'll need to test it, but I don't think empty structs will break things.

thatplguy (approved):

thatplguy:
I agree with your comment, that seems like a nice solution -- sorry, didn't see it before I reviewed.

We can also extend the OpenAPI <--> IR conversion to translate annotated Structs to additionalProperties.

liujed (approved):

PR #34, opened 2021-05-20
Add content type to context.

liujed:


visitors/http_rest/spec_visitor.go
```diff
@@ -486,13 +491,19 @@ func extendContext(cin Context, node interface{}) {
 				named = false
 			} else if x := meta.GetMultipart(); x != nil {
 				ctx.setValueType(BODY)
+				unknown := pb.HTTPBody_UNKNOWN.String()
+				contentType = &unknown
```
thatplguy, line 495: At some point, we'll want to do something smarter with multipart encodings, like track the different parts.  This is fine for now though.

thatplguy (approved):

visitors/http_rest/spec_visitor.go
liujed: Multipart metadata don't seem to have content types.

PR #35, opened 2021-05-20
Warn when there's a collision in name normalization

liujed:
Previously, we were panicking. Until the IR is fixed, best we can do is log collisions and arbitrarily choose one side of the collision.

### DIFF
--- a/visitors/http_rest/normalize_arg_names.go
+++ b/visitors/http_rest/normalize_arg_names.go
@@ -1,12 +1,12 @@
 package http_rest
 
 import (
-	"fmt"
 	"reflect"
 	"strconv"
 	"strings"
 
 	pb "github.com/akitasoftware/akita-ir/go/api_spec"
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 
 	. "github.com/akitasoftware/akita-libs/visitors"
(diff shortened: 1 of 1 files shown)


thatplguy (approved):

visitors/http_rest/normalize_arg_names.go
```diff
@@ -81,9 +81,11 @@ func (*argNameNormalizer) VisitDataChildren(self interface{}, c SpecVisitorConte
 }
 
 func (v *argNameNormalizer) setName(name ArgName) {
-	if _, ok := v.normalizationMap[name]; ok {
-		panic(fmt.Sprintf("Unexpected duplicated name for %v", name))
+	if existing, ok := v.normalizationMap[name]; ok {
+		glog.Warning("Non-normalized arg names ", existing, " and ", v.nonNormalizedArgName, " have conflicting normalized name ", name, ". Keeping the former and ignoring the latter.")
+		return
```
mgritter, line 86: Is this path only for conflicts? It looks like the warning may be inappropriate if there isn't any conflict, just the same name twice?

liujed: The same name twice _is_ a conflict.

PR #36, opened 2021-05-20
Convert strings in traces to other types depending on context

liujed:
When producing IR from traces, we don't always want to auto-convert strings to ints and bools. Doing this conversion in JSON data causes us to generate incorrect specs.


thatplguy (approved):

mgritter (approved):
This seems reasonable.

I was trying to find where we applied Formats or FormatKind (which I think would be a good place for annotating int-in-a-string or bool-in-a-string) but it looks like that is in superstar itself?

liujed:
I'm not sure where that is happening either, but Cole and I discussed formats briefly, and it wasn't immediately clear to us what the right thing to do for formats would be. This was the easiest short-term solution.

PR #37, opened 2021-05-21
Fix for when visiting starts at non-argument Data node

liujed:
An array-index-out-of-bounds issue occurred when visiting starts at a Data node that is not a top-level method argument.


liujed:
Merging for deployment to Manuel.

PR #38, opened 2021-06-01
Moving a type out of King's Cross

liujed:


mgritter (commented):
It seems like it would be better to convert the type in kings_cross/rest/common.go to an alias of this one.

I am not at all sure on why some structs are in api_schema and others are not?

thatplguy (approved):
Roughly speaking, types shared between the CLI and the backend are in `akita-libs`, but not types that only live in the backend.

PR #39, opened 2021-06-03
Ignore requests with 4xx responses during melding

liujed:
When melding a source `pb.Method` into a destination `pb.Method`, ignore requests in the source if the source has at least one response and all responses have 4xx response codes.


thatplguy (approved):
LGTM.

PR #41, opened 2021-06-07
Add a type for tag keys

liujed:
Introduced a `tags` package that acts as a central repository for all our tags.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -57,7 +57,7 @@ type CreateLearnSessionRequest struct {
 	BaseAPISpecRef *APISpecReference `json:"base_api_spec_ref,omitempty"`
 
 	// Optional key-value pairs to tag this learn session.
-	// We reserve tags with "x-akita" and "akita:" prefix for internal use.
+	// We reserve tags with "x-akita-" prefix for internal use.
 	Tags map[tags.Key]string `json:"tags,omitempty"`
 
 	// Optional name for the learn session.
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/akitasoftware/akita-libs/akid"
 	"github.com/akitasoftware/akita-libs/spec_summary"
+	"github.com/akitasoftware/akita-libs/tags"
 )
 
 // NetworkDirection is always relative to subject service.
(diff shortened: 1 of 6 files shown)


api_schema/schema.go
```diff
@@ -56,8 +57,8 @@ type CreateLearnSessionRequest struct {
 	BaseAPISpecRef *APISpecReference `json:"base_api_spec_ref,omitempty"`
 
 	// Optional key-value pairs to tag this learn session.
-	// We reserve tags with "x-akita" prefix for internal use.
-	Tags map[string]string `json:"tags,omitempty"`
+	// We reserve tags with "x-akita" and "akita:" prefix for internal use.
```
thatplguy, line 60: Did you decide to use `akita:` after all, or should we revert this comment?

tags/reserved_tags.go
```diff
@@ -0,0 +1,118 @@
...
+// Returns an error if the key is reserved for Akita internal use.
+func (k Key) CheckReserved() error {
+	if !k.IsReserved() {
+		return nil
+	}
+
+	return errors.New(`Tags starting with "x-akita-" are reserved for Akita internal use.`)
+}
+
+// Returns a map from parsing a list of "key=value" pairs.
```
thatplguy, line 97: Does this mean we only support one value per tag right now?  (It's not necessary to add multi-valued tags if we don't have them, but I don't know what the current tag implementation supports.)

thatplguy (approved):

tags/reserved_tags.go
liujed: This came from here: https://github.com/akitasoftware/akita-cli/blob/d68d57516102f4fef68f814eadf68d4aed08be7a/cmd/internal/tags/tags.go

api_schema/schema.go
liujed: Fixed in d858e0c.

tags/reserved_tags.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		k, v := Key(parts[0]), parts[1]
+		if _, ok := results[k]; ok {
+			return nil, errors.Errorf("tag with key %s specified more than once", k)
+		}
+
+		if err := CheckReservedKey(k); err != nil {
+			return nil, err
+		}
```
mgritter, line 121: I'm not fond of this (though I see it was in the original version) just because it means there's no back-door to set them in use cases we might not anticipate.

For the short term, this means we can't set the "source" tag to production because there's no CLI code that supports doing so (that I can find in the other PR.)  And we can't inject tags using the CLI for testing.

In the long term, this means that UI features depending on these tags are only supported for the integrations we have put into the CLI, and they aren't scriptable for other CI or test environments where they might be applicable.

I'd suggest removing the check entirely, or downgrading it to a warning.  Third-best would be an undocumented tags flag that does skip this check.

mgritter (approved):

thatplguy (approved):

tags/reserved_tags.go
liujed: In 276afc5.

PR #43, opened 2021-06-12
Extend `LearnSession` to include more information

liujed:
Introduced `ListedLearnSession` for use when listing traces. This is an extended version of `LearnSession` that includes the number of trace events and identifies the API models that were created from the trace.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -64,22 +64,97 @@ type CreateLearnSessionRequest struct {
 	Name string `json:"name"`
 }
 
-// Also used as a model in specs_db.
 type LearnSession struct {
-	tableName struct{} `pg:"learn_sessions"`
-
-	ID           akid.LearnSessionID `pg:"id,pk" json:"id"`
-	Name         string              `pg:"name" json:"name"`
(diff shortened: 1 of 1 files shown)


api_schema/schema.go
```diff
@@ -64,22 +64,97 @@ type CreateLearnSessionRequest struct {
...
+// presentation in the web console when listing learn sessions.
+// XXX Should inherit from LearnSession, but Go. :(
+type ListedLearnSession struct {
 	tableName struct{} `pg:"learn_sessions"`
 
-	ID           akid.LearnSessionID `pg:"id,pk" json:"id"`
-	Name         string              `pg:"name" json:"name"`
-	IdentityID   akid.IdentityID     `pg:"identity_id" json:"identity_id"`
-	ServiceID    akid.ServiceID      `pg:"service_id" json:"service_id"`
-	CreationTime time.Time           `pg:"creation_time" json:"creation_time"`
```
thatplguy, line 75: I'm not familiar with Go's ORM, but was the `pg:...` fragment intentionally removed here?

thatplguy: Never mind, I see how this has been structured.

thatplguy (approved):

PR #47, opened 2021-06-16
Add reserved tags for Kubernetes deployments

mgritter:
There are not many "generic" deployment tags here, I would welcome suggestions.

Unfortunately, any of the K8S info is something that will have to be part of the pod or Daemonset configuration for us to get it.  See https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/; there are no fixed environment variables.

I could come up with some tags specific to Heroku.

Probably we should have a hostname and host IP that are not tied to Kubernetes? Let me know what you think is the best approach here, since I didn't get it done in time to talk with either of you online.  :)

### DIFF
--- a/tags/reserved_tags.go
+++ b/tags/reserved_tags.go
@@ -57,6 +57,8 @@ const (
 
 	// Identifies the git commit hash from which the trace or spec was derived.
 	// Attached to traces or specs obtained from CI.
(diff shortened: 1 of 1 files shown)


liujed (approved):
Looks reasonable

PR #48, opened 2021-06-17
Add field-path element for representing a variant of a `OneOf`

liujed:
Previously, a one-of variant was represented by a `FieldName` named after the variant's hash. This is now replaced with a `OneOfVariant` that encodes how many variants there are, implicitly enumerates the variants, and identifies which variant is on the field path.

### DIFF
--- a/visitors/http_rest/field_path.go
+++ b/visitors/http_rest/field_path.go
@@ -16,6 +16,7 @@ type fieldPathElementKind int
 const (
 	fieldNameKind fieldPathElementKind = iota
 	arrayElementKind
+	oneOfVariantKind
 )
 
 type abstractFieldPathElement struct {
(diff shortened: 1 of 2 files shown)


mgritter (approved):
This is certainly better (and good enough to go out today.)  I'm not sure it's more meaningful to the user to put something here than to just omit it, though?

As I understand it from this morning's demo, the field used to be reported as

myField.<hash value>                UUID

with presumbably the other one-of type appearing later?

myField.<other hash value>      telephone number

So now it will look like

myField.(format 1 of 2)          UUID
myField.(format 2 of 2)          telephone number

?

thatplguy (approved):

thatplguy:
@mgritter yes, exactly.  It's not ideal -- we'll need to figure out a better way to show `oneof` alternatives in the future -- but it surfaces all the data we want to surface and doesn't look too horrible or confusing, hopefully.

It also addresses a second case, where the `oneof` includes more complicated objects.  In that case, you might have something like

```
myField.(format 1 of 2).subfield1.name       string
myField.(format 2 of 2).[].name              string
myField.(format 2 of 2).[].id                UUID
```

PR #49, opened 2021-06-17
Fix field path for one-of

liujed:
Need to look to grandparent to find the `OneOf`. The parent is just a `map[string]*Data`.

Actually count variants starting from 1.

### DIFF
--- a/visitors/visitor.go
+++ b/visitors/visitor.go
@@ -47,9 +47,14 @@ func (c ContextPath) GetLast() ContextPathElement {
 	return c[len(c)-1]
 }
 
-// Returns the `n`-th last element in the path.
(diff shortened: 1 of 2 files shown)


visitors/visitor.go
```diff
@@ -47,6 +47,11 @@ func (c ContextPath) GetLast() ContextPathElement {
 	return c[len(c)-1]
 }
 
+// Returns the `n`-th last element in the path.
+func (c ContextPath) GetNthLast(n int) ContextPathElement {
```
thatplguy, line 51: Can you add a check here and return `nil` if there's an array underflow?

thatplguy (approved):

visitors/visitor.go
liujed: In 94578a6.

PR #52, opened 2021-06-28
Add x-akita-truncated-spec

mgritter:


tags/reserved_tags.go
```diff
@@ -97,6 +97,9 @@ const (
 	// The name of the deployment environment, suggested values "production"
 	// or "staging" but may be a user-defined value
 	XAkitaDeployment Key = "x-akita-deployment"
+
+	// Used for specifications where the number of witnesses is too large.
+	XAkitaTruncatedSpec Key = "x-akita-truncated-spec"
```
liujed, line 102: Is the `-spec` redundant?

liujed (approved):
LGTM

tags/reserved_tags.go
mgritter: Hm...  yeah. The tag can't reasonably be applied to a trace (yet) but we don't include -spec for other spec-only tags.

PR #54, opened 2021-06-30
Add x-akita-created-by value for big models

liujed:


tags/created_by.go
```diff
@@ -6,4 +6,7 @@ type CreatedBy = string
 const (
 	// Designates a spec that was automatically created by a schedule.
 	CreatedBySchedule CreatedBy = "schedule"
+
+	// Designates a spec that was automatically created as a "big model".
+	CreatedByBigModel CreatedBy = "big model"
```
mgritter, line 11: This works, but my gut reaction was "can we avoid using a space character."

mgritter (approved):

tags/created_by.go
liujed: Replaced with `aggregator` in 6d35d03.

PR #57, opened 2021-07-09
Handle proprietary headers in spec summary.

mgritter:
First of the downstream changes handling the IR change in https://github.com/akitasoftware/akita-ir/pull/2.

Probably they need to go in together; I'll do the CLI changes next.


liujed (commented):
Here's another place where authorization types are read. I think this might also need to be adjusted. https://github.com/akitasoftware/akita-libs/blob/8388976a7962059b6e7c2f1fe60d0700f62d4e4a/visitors/http_rest/spec_visitor.go#L488-L493

mgritter:
> Here's another place where authorization types are read. I think this might also need to be adjusted.
> 
> https://github.com/akitasoftware/akita-libs/blob/8388976a7962059b6e7c2f1fe60d0700f62d4e4a/visitors/http_rest/spec_visitor.go#L488-L493

Do you think the context needs to be extended with the header name? Because as things stand, there's nothing more I can do in that case.

liujed:
I'm not sure what we should be doing. Maybe have `named` be true and set `name` to the name of the underlying header, now that it's not just "Authorization". We'd have to check all the places where this is used to make sure the name is no longer assumed to be "Authorization".

PR #58, opened 2021-07-13
Limit total HTTP capture size to 1MiB.

mgritter:
Tested by downloading a 1GB file, verified that HAR file contained only the first approximately 1MB of payload.

### DIFF
--- a/akinet/http/parser.go
+++ b/akinet/http/parser.go
@@ -14,9 +14,11 @@ import (
 	"github.com/akitasoftware/akita-libs/memview"
 )
 
-const (
-	// Terminate parsing on any request or response larger than 1MB.
-	maxHttpLength = 1024 * 1024
+var (
+	// Default maximum HTTP length supported.
+	// Can be altered by the CLI as a configuration setting, but doing so after parsing
+	// has started will be a race condition.
+	MaximumHTTPLength int64 = 1024 * 1024
 )
 
 // Implements TCPParser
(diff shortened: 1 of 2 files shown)


akinet/http/parser.go
```diff
@@ -14,6 +14,13 @@ import (
 	"github.com/akitasoftware/akita-libs/memview"
 )
 
+var (
+	// Default maximum HTTP length supported.
+	// Can be altered by the CLI as a configuration setting, but doing so after parsing
+	// has started will be a race condition.
+	MaximumHTTPLength int64 = 1024 * 1024
```
liujed, line 21: Is the thread/goroutine setting this is the same as the one reading? If not, I think the Go memory model asks for a synchronization point before the effects of the write are guaranteed visible.

We could add a parameter to `NewHTTP{Request,Response}ParserFactory` and thread the value through from there. Takes effort, though.

liujed (approved):
A concern about synchronization, but looks reasonable otherwise.

akinet/http/parser.go
liujed: I guess we're fine, since this will be set during processing of command-line options, before any goroutines are created.

mgritter: Yes, as long as it's done before the apidump.Run() command starts we should be fine.

akinet/http/parser_test.go
```diff
@@ -261,6 +261,7 @@ func TestHTTPRequestParser(t *testing.T) {
 	}
 
 	for _, c := range testCases {
+		t.Logf(c.name)
```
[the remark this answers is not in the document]
mgritter: Yeah, I thought one of the test cases was wedged but it was just taking a long time because we try O(n^2) different splits.  The actual error messages include the test name, but I still find it helpful to be able to see which test case has started.

PR #59, opened 2021-07-16
Support for Akita model versions

thatplguy:
Model versions are like tags, but unique across models within the same service.
This PR adds support for detecting version names reserved by Akita, as well as
a short README explaining the difference between model versions and tags.


model_versions/reserved_versions.go
```diff
@@ -0,0 +1,13 @@
...
+
+const (
+	XAkitaLatestModelVersion = "latest"
+)
+
+// Determines whether a version is reserved for Akita internal use.
+func IsReservedModelVersion(k string) bool {
+	s := strings.ToLower(string(k))
+	return strings.EqualFold(s, XAkitaLatestModelVersion)
+}
```
mgritter, line 13: We also use the x-akita-big-model prefix for internal versions.

https://github.com/akitasoftware/superstar/blob/467ec50cac624b4c1b7a44d7615ef3d8a6e740e5/lib/versions/versions.go#L15

mgritter (approved):

mgritter (changes requested):
On further thought, this should be merged with https://github.com/akitasoftware/superstar/blob/467ec50cac624b4c1b7a44d7615ef3d8a6e740e5/lib/versions/versions.go one way or the other.

thatplguy:
Good call.  Since this needs to be shared across the CLI and the backend, I'll factor the other version info into here.

mgritter (approved):

PR #60, opened 2021-07-16
Remove 'stable' from the list of reserved version names

thatplguy:


mgritter (approved):
Sorry, should have caught this on review :)

thatplguy:
Yea, I felt a little silly when I went to test the "mark as stable" workflow -- the whole reason for doing this! -- and realized it would't work :)

PR #61, opened 2021-07-23
Initial support for inferring maps

liujed:
When melding IR structures, if we end up with a struct with more than 15 optional fields or 100 required fields, convert it into a map. Currently, map keys are just unformatted strings. Inferring formats of map keys is left for a future PR.


liujed:
Depends on akitasoftware/akita-ir#3.

spec_util/testdata/meld/meld_map_1_map_2_expected.pb.txt
```diff
@@ -0,0 +1,202 @@
...
+              options: <
+                key: "9v8pKk6Y-tY="
+                value: <
+                  primitive: <
+                    string_value: <
+                      value: "addr1@example.com"
+                    >
+                  >
+                >
+              >
```
liujed, line 23: Need to
1. Infer that these are e-mail addresses and
2. Remove the concrete values.

Where in our code do we do this?

mgritter: 1. Is done in the client today, the library is superstar/lib/data_formats (GetFormats)
2. happens in lib/learn/sanitize.go with the sanitizingHasher

spec_util/meld.go
```diff
@@ -446,15 +462,113 @@ func meldStruct(dst, src *pb.Struct) error {
...
 
-	// ... but what if we end up with hundreds of keys?
-	// Probably that means the key is part of the data. Let's use an arbitrary
-	// threshold, and if we cross it, merge all the keys.
-	// TODO: add another field to Struct in in the IR specifying that this has occurred.
-	// (and if we can infer a data format on the keys, it could live at that level too.)
+	// Apply a heuristic for deciding when to convert structs to maps.
+	if structShouldBeMap(dst) {
+		structToMap(dst)
+	}
```
mgritter, line 468: I've been concerned about the cost of this entire method (and about the cost of re-hashing the structure once it is done!)  And the additional loop does not help.

A small improvement would be to track whether we have made any changes or not, and skip this heuristic check (and, ideally, the re-hashing) if we have not.

mgritter (commented):
Looks pretty good, waiting on format inference and value removal.

spec_util/testdata/meld/meld_map_1_map_2_expected.pb.txt
liujed: Hard-coding map keys as unformatted strings for now (9892d00).

spec_util/meld.go
```diff
@@ -446,15 +462,113 @@ func meldStruct(dst, src *pb.Struct) error {
...
 	return nil
 }
 
+// Determines whether the given pb.Struct represents a map.
+func isMap(struc *pb.Struct) bool {
+	return struc.MapType != nil
+}
+
+// Tuning parameter for deciding when a struct should be turned into a map.
+const maxOptionalFieldsPerStruct = 15
```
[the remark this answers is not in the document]
liujed: Set to 100 in 9892d00. Should it be higher?

spec_util/meld.go
liujed: Won't be addressed in this PR.

liujed:
This PR is still WIP. At least need to update visitor contexts to represent this new kind of map context.

liujed:
I think this is done, modulo reviews. Will wait for integration into superstar before landing this.

mgritter (approved):
Looks OK.

PR #62, opened 2021-07-28
Reduce amount of hashing for methods

liujed:
If IR trees are consistent, then the requests and responses for each `Method` instance should be mapped by their hashes. When hashing a `Method`, use these hash values instead of recomputing them.

### DIFF
--- a/spec_util/testdata/generalize_witnesses/spec.pb.txt
+++ b/spec_util/testdata/generalize_witnesses/spec.pb.txt
@@ -63,7 +63,7 @@ methods: <
     >
   >
   args: <
-    key: "y0HmMXfg67c="
+    key: "NG5nItAmli0="
     value: <
       primitive: <
         int64_value: <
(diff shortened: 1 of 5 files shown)


liujed:
Depends on akitasoftware/objecthash-proto#1.

thatplguy (approved):

spec_util/testdata/generalize_witnesses/spec.pb.txt
```diff
@@ -1474,7 +1474,7 @@ methods: <
     >
   >
   args: <
-    key: "jX1cqTk1FwI="
+    key: "Sajk40aPGIE="
```
thatplguy, line 1477: These are changing because they were previously incorrect, I'm assuming?  I know I've run into incorrect hashes in the unit test data before.

liujed: Yes, they were incorrect. Pretty much why I added the sanity check to the test harness.

PR #63, opened 2021-07-28
Reserved tags for specs copied to support.

mgritter:


tags/reserved_tags.go
```diff
@@ -147,6 +147,19 @@ const (
 	XAkitaDumpFilterFlag Key = "x-akita-dump-filter-flag"
 )
 
+// Tags applied to a copy of the spec
+const (
+	XAkitaOriginalOrganization = "x-akita-original-organization"
```
liujed, line 152: Is this an ID or a name?

tags/reserved_tags.go
```diff
@@ -147,6 +147,19 @@ const (
...
+
+	XAkitaOriginalService = "x-akita-original-service"
```
liujed, line 154: `XAkitaOriginalServiceName = "x-akita-original-service-name"`

tags/reserved_tags.go
```diff
@@ -147,6 +147,19 @@ const (
...
+
+	XAkitaOriginalServiceID = "x-akita-original-service-id"
+
+	XAkitaOriginalName = "x-akita-original-name"
```
liujed, line 158: `XAkitaOriginalSpecName = "x-akita-original-spec-name"`

liujed (approved):
Nits for some slightly less ambiguous tag names.

thatplguy (approved):

tags/reserved_tags.go
mgritter: Changed to use ID suffix to be consistent.

tags/reserved_tags.go
mgritter: I changed to x-akita-original spec; original-service and original-spec are names; original-service-id and original-spec-id are IDs.

PR #66, opened 2021-07-31
Rewrote Memview.Index

mgritter:
Added benchmark tests.
Added a failing test not relevant to our actual usage.

Old implementation:
```
BenchmarkIndexSmall-8   	   94458	     59803 ns/op
BenchmarkIndexLarge-8   	     337	  17548862 ns/op
```

New implementation:
```
BenchmarkIndexSmall-8   	  435463	     12523 ns/op
BenchmarkIndexLarge-8   	    1570	   3629018 ns/op
```

This could be improved further by searching in parallel for a set of strings; the benchmark roughly reflects how we actually use it.

### DIFF
--- a/memview/memview.go
+++ b/memview/memview.go
@@ -150,11 +150,20 @@ func (mv MemView) Index(start int64, sep []byte) int64 {
 
 	// Iteratively search for the target, keeping in mind that the target may be
 	// spread over multiple slices in mv.buf.
+	//
+	// TODO: this only works correctly for search strings that do not have a repeated
+	// prefix. To work correctly, we would have to back up to the point at which
+	// the needle *could* have started after an incomplete match.
+	//
+	// However, we only use this method to search for strings without a repeated prefix:
+	// GET, POST, DELETE, HEAD, PUT, PATCH, CONNECT, OPTIONS, TRACE, HTTP/1.1 and HTTP/1.0
 	needle := sep
 	needleIndex := 0
 	for b := startBuf; b < len(mv.buf); b++ {
 		haystack := mv.buf[b]
-		for i := startOffset; i < len(haystack); i++ {
+		// Check remainder of needle if overlap from last buffer
+		var i int = 0
+		for i = startOffset; i < len(haystack) && needleIndex > 0; i++ {
 			if haystack[i] == needle[needleIndex] {
 				needleIndex += 1
 				if needleIndex == len(needle) {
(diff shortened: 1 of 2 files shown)


mgritter:
We are generally in a situation where our buffers are large and our search strings are small.  So it's probably worth using bytes.Index() which has a platform-optimized implementation, and then this checking the len(needle) wraparound cases at each boundary.

That would also solve the correctness problem.

However, it's probably still less efficient than parallel search for the set of strings we're interested in.

memview/memview.go
```diff
@@ -151,50 +151,26 @@ func (mv MemView) Index(start int64, sep []byte) int64 {
...
+				if needleIndex == len(needle) {
+					// Found, figure out start index.
+					// At the start of the 'i' loop, it points to currentIndex, so we
+					// need to add i and subtract startOffset.  Then move back to the
+					// first character in the needle
+					return currIndex + int64(i-startOffset) - int64(len(needle)-1)
 				}
-				break
+			} else {
+				needleIndex = 0
```
thatplguy, line 168: Don't we need to reset the haystack pointer as well?  I think there's a corner case where the search fails to find the needle:
```
haystack: aaa_foo
needle: aa_foo

When i == 2, needleIndex == 2, and haystack[i] != needle[needleIndex].
When i == 3, needleIndex == 0, and we've missed finding the needle (found when i == 1 and needleIndex == 0).
```

thatplguy: Ah hah, I see this is the test case you added below 🙂

thatplguy (approved):

memview/memview.go
mgritter: Yes, I have faithfully reproduced the existing bugs.  :)

mgritter:
New benchmarking:

Previous version in PR:
```
BenchmarkIndexSmall-8   	 3002158	     11870 ns/op
BenchmarkIndexLarge-8   	    9596	   3624710 ns/op
```

Most recent:
```
BenchmarkIndexSmall-8   	12983755	      2619 ns/op
BenchmarkIndexLarge-8   	   23590	   1597876 ns/op
```

This is by using `bytes.Index()` instead of rolling our own.

I tried a version of parallel search using `regexp` but it was much slower (about 10000/30 seconds instead of 23590) so that is not included in the PR.

mgritter:
Re-requested from @thatplguy due to significant changes.

thatplguy (approved):
LGTM

PR #69, opened 2021-08-03
Fix GetInnermostNode(), add a test for normalization that covers it.

mgritter:
Wrote a test that reproduces the bug found in superstar unit testing.
Fixed GetInnermostNode().
Verified that there are no other methods that use .path or GetPath()


mgritter:
Will try superstar unit tests.

thatplguy (approved):

PR #71, opened 2021-08-06
Lightweight hashing for Akita IR

mgritter:
This PR has a code generator (gen.go) that outputs Hash<X> functions for the Akita IR which exactly match the results of objecthash-proto.  But, the resulting code avoids reflection and temporary buffers.

The two things that the generated code does better than the original code are:
1. Process struct fields in the order in which they would be sorted (based on the hash of their integer tag field) instead of sorting afterwards.
2. Send intermediate results directly to the hash function instead of creating temporary buffers.  (Which we can do for structs because we know the order in which they need to be hashed.)

This appears successful at lowering the amount of memory allocated:

```
BenchmarkWitnessHash-8      	   15476	     76078 ns/op	   18349 B/op	     947 allocs/op
BenchmarkWitnessOldHash-8   	    7077	    173922 ns/op	   48664 B/op	    1561 allocs/op
```

generated_types.go is the generated file, which can be regenerated with the provided Makefile.


thatplguy (approved):
LGTM.  Can you update the README instructions in `akita-ir` to indicate that new structs added to the IR should also be added here?

spec_util/ir_hash/wrappers.go
```diff
@@ -0,0 +1,17 @@
+package ir_hash
+
+import (
+	"encoding/base64"
+
+	pb "github.com/akitasoftware/akita-ir/go/api_spec"
+)
+
+func HashWitnessToString(w *pb.Witness) string {
```
thatplguy, line 9: Hashing can't fail?  What would happen if I changed the IR but didn't change this file in the future, and then tried to hash a witness with added/removed fields?

mgritter: Because it's not using reflection, there is no way to know whether additional fields are present or not, so it would silently exclude them from the hash.

It would be nice if we could attach some version number to the generated go objects (that could be easily checked)--- let me see if we can do that with fileDescriptor.

mgritter: https://github.com/akitasoftware/akita-libs/pull/71/commits/0d71d81eb46c8c9108929d300700909e915b9778 stores hashes of the protobuf "file descriptors", which should pick up any added types and cause a panic during tests.

PR #75, opened 2021-08-19
Increase threshold for converting to maps

liujed:
Now converting object types to maps when there are more than 50 optional fields or 100 required fields.

### DIFF
--- a/spec_util/testdata/meld/meld_map_2.pb.txt
+++ b/spec_util/testdata/meld/meld_map_2.pb.txt
@@ -9,7 +9,7 @@ method {
     value {
       struct {
         fields {
-          key: "addr9@example.com"
+          key: "addr26@example.com"
           value {
             primitive {
               string_value {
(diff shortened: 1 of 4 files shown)


spec_util/testdata/meld/meld_map_2.pb.txt
```diff
@@ -72,7 +72,169 @@ method {
...
+          key: "addr50@example.com"
+          value {
+            primitive {
+              string_value {
+              }
+            }
+          }
+        }
+        fields {
+          key: "addr51@example.com"
```
mgritter, line 237: I think that meld_map_3.pb.txt, which was a copy of meld_map_2.pb.txt with one type change, would no longer exhibit the bug after the threshold change?

mgritter (approved):

spec_util/testdata/meld/meld_map_2.pb.txt
liujed: Good catch. Forgot to update main when I created the branch.

liujed: Fixed in 38df7fe.

PR #82, opened 2021-09-25
Make interval type explicit.

mgritter:
TimeSpan is now an alias for HalfOpenInterval.
I ended up using HalfOpenInterval so the others may be less well-developed.

### DIFF
--- a/time_span/time_span.go
+++ b/time_span/time_span.go
@@ -39,11 +39,15 @@ func NewClosedInterval(start time.Time, end time.Time) ClosedInterval {
 	}
 }
 
+func (span ClosedInterval) Empty() bool {
+	return span.End.Before(span.Start)
+}
+
 func (span ClosedInterval) Duration() time.Duration {
-	if span.End.After(span.Start) {
-		return span.End.Sub(span.Start)
-	} else {
+	if span.Empty() {
 		return time.Duration(0)
+	} else {
+		return span.End.Sub(span.Start)
 	}
 }
 
@@ -53,10 +57,18 @@ func (span ClosedInterval) Includes(query time.Time) bool {
 }
 
 // Extend an interval by "delta" in each direction (thus it gets 2*delta longer.)
+// If it was empty, expand around the start point.
 func (t ClosedInterval) Expand(delta time.Duration) ClosedInterval {
-	return ClosedInterval{
-		Start: t.Start.Add(time.Duration(-1) * delta),
-		End:   t.End.Add(delta),
+	if t.Empty() {
+		return ClosedInterval{
+			Start: t.Start.Add(time.Duration(-1) * delta),
+			End:   t.Start.Add(delta),
+		}
+	} else {
+		return ClosedInterval{
+			Start: t.Start.Add(time.Duration(-1) * delta),
+			End:   t.End.Add(delta),
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


time_span/time_span.go
```diff
@@ -2,28 +2,210 @@ package time_span
...
+
+// Return the intersection of two half-open intervals.
+// (i.e., trim this interval to lie entirely within another.)
+// May result in an interval with end <= start, which should be treated as empty.
+func (t HalfOpenInterval) Intersect(t2 HalfOpenInterval) HalfOpenInterval {
+	return HalfOpenInterval{
+		Start: MaxTime(t.Start, t2.Start), // later of the two starts
+		End:   MinTime(t.End, t2.End),     // earlier of the two ends
+	}
+}
```
liujed, line 189: Again, issue with empty intersections.

time_span/time_span.go
```diff
@@ -2,28 +2,210 @@ package time_span
...
 }
 
-func (span TimeSpan) Duration() time.Duration {
-	return span.End.Sub(span.Start)
+func (span ClosedInterval) Duration() time.Duration {
+	if span.End.After(span.Start) {
+		return span.End.Sub(span.Start)
+	} else {
+		return time.Duration(0)
+	}
```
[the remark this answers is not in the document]
mgritter: Having no away to represent empty intervals also seems surprising.

If Includes does not work for empty intervals, I'll fix it, but I'd rather that Intersect returns empty intervals than that it return an error.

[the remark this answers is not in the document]
liujed: If we're letting `Start > End` represent the empty intersection, that's fine, but let's document this. Rather than carefully thinking about whether our operations work for the cases where `Start > End`, what if we introduced an `IsEmpty()` and checked this case explicitly? I still think that `Expand` needs to be able to return an error, though.

time_span/time_span.go
```diff
@@ -2,28 +2,210 @@ package time_span
...
+	// There's some time C such that start1 <= C < end1 and start2 <= C < end2
+	// iff start1 < end2 AND start2 < end1.
+	return t1.Start.Before(t2.End) && t2.Start.Before(t1.End)
+}
+
+// Returns true if t2 is entirely contained within t1
+func (t1 HalfOpenInterval) Contains(t2 HalfOpenInterval) bool {
+	return (t1.Start.Before(t2.Start) || t1.Start.Equal(t2.Start)) &&
+		(t1.End.After(t2.End) || t1.End.Equal(t2.End))
+}
```
liujed, line 171: Passing different representations of the empty interval to this implementation can yield different results:
```
nonEmpty := [1, 2)
empty1 := [2, 1)
empty2 := [3, 0)
empty3 := [4, 3)
nonEmpty.Contains(empty1) // returns true
nonEmpty.Contains(empty3) // returns false
empty1.Contains(empty2) // returns true
empty1.Contains(empty3) // returns false
```

time_span/time_span.go
```diff
@@ -2,28 +2,210 @@ package time_span
...
+
+// Extend an interval by "delta" in each direction (thus it gets 2*delta longer.)
+func (t HalfOpenInterval) Expand(delta time.Duration) HalfOpenInterval {
+	return HalfOpenInterval{
+		Start: t.Start.Add(time.Duration(-1) * delta),
+		End:   t.End.Add(delta),
+	}
+}
```
liujed, line 179: Similar issue as for closed intervals.

time_span/time_span.go
```diff
@@ -2,28 +2,210 @@ package time_span
...
+
+// Return the "union" of two half-open intervals, along with everything in between.
+// TOOD: I don't know if this has a nice mathematical name.
+func (t HalfOpenInterval) Combine(t2 HalfOpenInterval) HalfOpenInterval {
+	return HalfOpenInterval{
+		Start: MinTime(t.Start, t2.Start), // earlier of the two starts
+		End:   MaxTime(t.End, t2.End),     // later of the two ends
+	}
+}
```
liujed, line 198: Combining two empty intervals can result in a non-empty interval:
```
empty1 := [1, 0)
empty2 := [3, 2)
empty1.Combine(empty2) == [1, 2)
```
Similarly, combining an empty interval with a non-empty interval shouldn't result in a different non-empty interval.
```
empty := [1, 0)
nonEmpty := [2, 3)
nonEmpty.Combine(empty) == [1, 3)
```

liujed (commented):
Modified my review to account for our letting `Start > End` represent empty intersections.

liujed:
It occurs to me that users of this API might be surprised to find `Start > End` when they directly access `Start` and `End` for intervals that represent empty intersections.

mgritter:
> It occurs to me that users of this API might be surprised to find `Start > End` when they directly access `Start` and `End` for intervals that represent empty intersections.

No more surprised, presumably, than if they were not using the library?

I mean, let's review: you suggested that the functions I was using be put in a library, and clarify whether we were dealing with open or closed intervals.  Now we have a bunch of extra requirements to make a sensibly-behaving library. But either the client code is prepared to deal with an intersection being empty, or is not, and nothing I can put in the library will force them do to so, except to go the sum type route of defining an empty interval as its own structure.

Having the intersection return error can't be the right choice, although a fixed representation of empty intervals might be (but could be more surprising, when the values given in the constructor are not the ones used.)

time_span/time_span.go
mgritter: Right, the problem is that this operation isn't a union, it's something weird.

The use case is: we have an event at [0,1) and we have an event at [4,5) and we want to merge it to a single interval.

If we have [0,0) instead then I would still want to merge to [0,5) even though the former is empty.

If we have [1,0) somehow then I am actually fine with [1,5) as the combination. I don't much care.

I would be happy to take this back out of the library so I can get the semantics I want for the combination, rather than arguing about what it would be in general.

liujed:
> I mean, let's review: you suggested that the functions I was using be put in a library, and clarify whether we were dealing with open or closed intervals. Now we have a bunch of extra requirements to make a sensibly-behaving library. But either the client code is prepared to deal with an intersection being empty, or is not, and nothing I can put in the library will force them do to so, except to go the sum type route of defining an empty interval as its own structure.

I was expecting a quick-and-dirty library that was private to where it is needed. That this is being offered as part of `akita-libs` means it ought to be resilient to someone picking this up later and using it without understanding its internals. This particular concern regarding `Start` and `End` arises because directly accessing those fields breaks the abstraction. One option is to rename these fields `start` and `end` so that they are private, and provide accessors that return an error in the cases where the interval's start or end is not well defined. Not the prettiest, but such is the way of Go.

time_span/time_span.go
liujed: I'd advocate more for clear documentation than carefully figuring out what the semantics needs to be. For the cases where we are uncertain or don't currently care, we can just say that the result is undefined.

liujed (approved):
LGTM. I've made a few tweaks in #83.

PR #84, opened 2021-09-29
Cleaner timeline API

liujed:
Omit values in `api_schema.Timeline` instead of using `"*"`.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -362,13 +362,23 @@ const (
 
 )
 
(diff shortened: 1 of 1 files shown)


mgritter (approved):
Looks good; this can then be re-used for the examples or other endpoint keys.

Should check with Cole how he feels about "" vs "*" in the response, "" seems fine and we can present it however we want.

thatplguy (approved):
LGTM.  Switching from `*` to optional fields is fine by me.

PR #88, opened 2021-10-20
Upgrade protobuf to v2 go api.

mgritter:
Adjust code-generation to match.
Upgrade to matching akita-ir.
Switch to using protocmp for go-cmp of protobufs.


spec_util/ir_hash/gen/gen.go
```diff
@@ -32,13 +32,17 @@ func main() {
 	gf.AddIgnoredField("HTTPMethodMeta", "ProcessingLatency")
 	gf.AddIgnoredField("Data", "ExampleValues")
 
+	// Newer protobuf compiler uses these istead of XXX prefix
```
thatplguy, line 35: Nit: Might be nice to put the old vs. new versions in the comment, e.g. `Newer protobuf compiler (1.39) uses these instead of XXX prefix (1.26)`

liujed (approved):

PR #91, opened 2021-11-06
Fix for go-pg

liujed:


mgritter (approved):
Hmm... this is this necessary, doesn't our binding for LearnSessionID do the right thing?

liujed:
> Hmm... this is this necessary, doesn't our binding for LearnSessionID do the right thing?

Explained here: https://github.com/akitasoftware/superstar/pull/518#issuecomment-962464366.

thatplguy (approved):

PR #93, opened 2021-11-09
Change DefaultSpecVisitor to avoid reflection.

mgritter:
New method method uses custom code for each type.
(Finally got it simple enough that it *could* be automated.)

### DIFF
--- a/visitors/http_rest/spec_visitor_default.go
+++ b/visitors/http_rest/spec_visitor_default.go
@@ -94,7 +94,6 @@ func DefaultVisitIRChildren(ctx Context, vm VisitorManager, m interface{}) Cont 
 		}
 	case *pb.Data_Struct:
 		if node != nil {
-
 			keepGoing = visitStructMembers(ctx, vm, m, "Struct", node.Struct)
 		}
 	case *pb.Data_List:
--- a/visitors/http_rest/spec_visitor_default.go
+++ b/visitors/http_rest/spec_visitor_default.go
@@ -1,0 +1,316 @@
+package http_rest
+
+import (
+	"fmt"
+
+	pb "github.com/akitasoftware/akita-ir/go/api_spec"
+	. "github.com/akitasoftware/akita-libs/visitors"
+	"github.com/akitasoftware/akita-libs/visitors/go_ast"
+)
+
+// This is the default implementation of VisitNodeChildren for DefaultSpecVisitorImpl.
+// It is indirectly called by every Visit____Children that does not override the default method.
+//
+// We need to call the visitor manager on:
+//   Every exported field in a struct, after updating the context with EnterStruct.
+//   Every element of an array, after updating the context with EnterArray.
+//   Every element of a map, after updating the context with EnterMap.
+// Each call to visit may abort the operation.
+//
+// This implementation uses reflection as a fallback, but has specialized methods for common types.
+// GRPC types use the fallback.
+//
+// It looks like the original implementation would call EnterNode/VisitChildren/LeaveNode on basic types.
+// We won't be doing that, as specVisitor's enter/visitChildren/leave methods do not do anything with them.
+//
+func DefaultVisitIRChildren(ctx Context, vm VisitorManager, m interface{}) Cont {
+	keepGoing := Continue
+
+	// A check for m == nil doesn't work here, because m could be a typed pointer
+	// to nil, which is different than interface{}(nil).
+
+	switch node := m.(type) {
+	case *pb.APISpec:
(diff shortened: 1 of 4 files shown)


visitors/http_rest/spec_visitor_default.go
```diff
@@ -0,0 +1,315 @@
...
+	default:
+		// Fallback to reflection-based implementation
+		return go_ast.DefaultVisitChildren(ctx, vm, m)
+	}
+	return keepGoing
+}
+
+// Visit the struct members given by <field1> <value1> <field2> <value2>
+func visitStructMembers(ctx Context, vm VisitorManager, inStruct interface{}, fields ...interface{}) Cont {
+	keepGoing := Continue
```
mgritter, line 218: I initially wrote a version that just took one field and one value.  And then I decided that was silly, so I made this variadic.  But now I wonder whether Go is actually going to allocate a slice on the heap on every call?  Hopefully not.

visitors/http_rest/spec_visitor_default.go
```diff
@@ -0,0 +1,315 @@
...
+func DefaultVisitIRChildren(ctx Context, vm VisitorManager, m interface{}) Cont {
+	keepGoing := Continue
+
+	// A check for m == nil doesn't work here, because m could be a typed pointer
+	// to nil, which is different than interface{}(nil).
+
+	switch node := m.(type) {
+	case *pb.APISpec:
+		// Methods is a []*Method, but Tags is just map[string]string
+		keepGoing = visitStructMembers(ctx, vm, m, "Methods", node.Methods)
```
liujed, line 35: This code belongs in `DefaultSpecVisitorImpl`'s implementation of `VisitAPISpecChildren`. Similarly for many of the other cases in this switch: `VisitNodeChildren` should just be a fallback for node types that don't have a more specific visit-children method.

As an aside, we really should have a set of enter/visitChildren/leave methods for all AST node types (including, e.g., Data_Struct) , but a lot of these node types are crufty protobuf artifacts, and I was too lazy to put in visitor methods for them. In any case, the whole visitor framework should just be generated from the IR, to ensure consistency with changes in the IR.

visitors/http_rest/spec_visitor_default.go
```diff
@@ -0,0 +1,315 @@
...
+	default:
+		panic(fmt.Sprintf("Unknown Cont value: %d", keepGoing))
+	}
+}
+
+// Visits the node m of any type, whose context is c. This should never return SkipChildren.
+// Implements all the logic to enter, visit children, and leave, while obeying the visitor's returned
+// Cont value.
+// This is a copy of astVisitor's visit function.
+func VisitInterface(c Context, vm VisitorManager, m interface{}) Cont {
```
liujed, line 274: As the comment says, this is basically just `visit()` lifted to the `VisitorManager` level. I never really understood why we need both visitors and `VisitorManager`. Seems like we should eventually just merge the two together.

liujed (approved):
Looks okay, though some of the code in `DefaultVisitIRChildren` belongs in the visit-children methods defined in `DefaultSpecVisitorImpl`.

visitors/http_rest/spec_visitor_default.go
mgritter: Hm.... OK, I guess I need to talk with you about this, because that's why I was asking about it on Slack:

> Mark Gritter  9:16 PM
> Is the intention in DefaultSpecVisitorImpl that a visitor could provide its own VisitNodeChildren, that would take over as the default for every other type?  I think that's what you meant, but I can't find any example of actually doing that.
> So a replacement for go_ast.DefaultVisitChildren(ctxt, vm, node) would have to be a function there, leaving all the Visit*Children default implementations unchanged?
> 
> Jed Liu  9:19 PM
> Yes, that was the intent. Currently unused, as you've observed.

liujed: These are mutually consistent. I understood the Slack question to be about how to override the default visiting behaviour for a generic `Node` type. I didn't realize that you intended to override it with something that inspected the dynamic type of the node. Avoiding this kind of type inspection is part of the intent behind the visitor framework.

visitors/http_rest/spec_visitor_default.go
liujed: Deduplicated as part of #94.

visitors/http_rest/spec_visitor_default.go
liujed: Refactored in #94.

PR #94, opened 2021-11-09
Refactored

liujed:
Moved some functionality out of `DefaultVisitIRChildren`. Functionality specific to node types for which we have a visit-children method belongs in that method.

Deduplicated code in `VisitInterface`.

### DIFF
--- a/visitors/http_rest/spec_visitor.go
+++ b/visitors/http_rest/spec_visitor.go
@@ -137,7 +137,8 @@ func (*DefaultSpecVisitorImpl) EnterAPISpec(self interface{}, c SpecVisitorConte
 }
 
 func (*DefaultSpecVisitorImpl) VisitAPISpecChildren(self interface{}, c SpecVisitorContext, vm VisitorManager, spec *pb.APISpec) Cont {
-	return self.(DefaultSpecVisitor).VisitNodeChildren(self, c, vm, spec)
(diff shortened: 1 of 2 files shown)


visitors/http_rest/spec_visitor.go
```diff
@@ -347,7 +381,13 @@ func (*DefaultSpecVisitorImpl) EnterPrimitive(self interface{}, c SpecVisitorCon
 }
 
 func (*DefaultSpecVisitorImpl) VisitPrimitiveChildren(self interface{}, c SpecVisitorContext, vm VisitorManager, d *pb.Primitive) Cont {
-	return self.(DefaultSpecVisitor).VisitNodeChildren(self, c, vm, d)
+	if d != nil {
+		return visitStructMembers(c, vm, d,
+			"Value", d.Value,
+			"AkitaAnnotations", d.AkitaAnnotations, // TODO: don't recurse into this one?
```
mgritter, line 387: Given that we don't have a VisitAkitaAnnotations, should we not recurse into this value?

mgritter (approved):

visitors/http_rest/spec_visitor.go
liujed: We probably should have a `VisitAkitaAnnotations`. Maybe some day, when we auto-generate this code.

PR #95, opened 2021-11-09
Various fixes

liujed:
* Added missing cases for IR node types in `DefaultVisitIRChildren`.
* Added missing case in `visitChildren`.
* `go mod tidy`


mgritter (approved):
Oh, HttpMeta _was_ missing, just from the other list!

PR #99, opened 2021-11-13
Remove `Outbound` from `NetworkDirection`

liujed:
It is no longer used.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -13,8 +13,7 @@ import (
 type NetworkDirection string
 
 const (
-	Inbound  NetworkDirection = "INBOUND"
-	Outbound NetworkDirection = "OUTBOUND"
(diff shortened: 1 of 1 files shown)


mgritter (approved):
I'm a little unsure about this one, but I'm willing to go along.

(Why does the type still exist if it has only one value?)

liujed:
The type is being kept for compatibility with older clients; it's still referenced here: https://github.com/akitasoftware/akita-libs/blob/c19e00f8027b158b7215d32f0cd7b272517201bb/api_schema/schema.go#L210-L213

liujed:
> (I'm assuming that the removal of `"OUTBOUND"` won't cause deserialization errors for "outbound" witnesses. I should check this...)

Tested here: https://go.dev/play/p/AxryI_sI9My.

PR #100, opened 2021-11-18
TLS support

liujed:
This parses TLS 1.2 and 1.3 Client Hello, Server Hello, and server certificate messages. From the client, we extract the hostname from SNI and a list of protocols from ALPN. From the server, we extract the TLS version. For TLS 1.2, we also extract the selected protocol from ALPN and the server's hostnames from the certificate's SAN extension.


akinet/tls/client_parser_factory.go
```diff
@@ -0,0 +1,64 @@
...
+
+		// Client Version (2 bytes)
+		9: 0x03, 10: 0x03, // protocol version 3.3 (TLS 1.2)
+	}
+
+	for idx, expectedByte := range expectedBytes {
+		if input.GetByte(int64(idx)) != expectedByte {
+			return akinet.Reject, input.Len()
+		}
+	}
```
mgritter, line 57: OK, I see what you did there.

I think the first problem is that expectedBytes should be at package level. Go will create a new map on each call to parse, when we should reuse the same one. 

I did a benchmark: https://gist.github.com/mgritter/62157436f582569add1ea24b5088ba22
234ns for creating the map inside the loop, vs. 73ns for outside

(Similar results for having it be a package-level var.)

We can get an even bigger win by masking a byte array instead of a map:

```
	expected := []byte{0x16, 0x03, 0x01, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x03, 0x03}
	mask := []byte{0xff, 0xff, 0xff, 0x00, 0x00, 0xff, 0x00, 0x00, 0x00, 0xff, 0xff}

	for n := 0; n < b.N; n++ {
		result = true
		for i := range expected {
			if (mv.GetByte(int64(i)) & mask[i]) != expected[i] {
```

That gets the cost of a successful check down to 12-13 ns.  Which seems worth it because this check is run every time we have unknown bytes we're trying to determine a parser for?

mgritter (approved):
This looks robust but I'd love to see some unit tests.

There is a big performance win in switching from a map of bytes to a masked array of bytes. Even if we don't go that far, moving the literals outside of the Accepts call seems worthwhile and easy.

I'm happy to go with seek() as-is and get some performance results, but it feels like it's a lot of memory pressure and overhead compared with starting up a Reader.  (And maybe adding a seek method to Reader.)

akinet/tls/util.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/akitasoftware/akita-libs/memview"
+)
+
+// Returns the uint16 at buf[offset..offset+1]. Returns 0 if offset+1 >=
+// len(buf).
+func uint16At(buf memview.MemView, offset int64) uint16 {
+	if buf.Len() <= offset+1 {
+		return 0
+	}
+	return binary.BigEndian.Uint16([]byte(buf.SubView(offset, offset+2).String()))
```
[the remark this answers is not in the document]
liujed: Moved into `MemView` in eb92ea1.

akinet/tls/client_parser_factory.go
liujed: In 3d7e5e9.

akinet/tls/client_parser.go
```diff
@@ -0,0 +1,242 @@
...
+	// Wait until we have the full handshake record.
+	if parser.allInput.Len() < handshakeMsgEndPos {
+		return nil, 0, nil
+	}
+
+	// Get a Memview of the handshake record.
+	buf := parser.allInput.SubView(tlsRecordHeaderLength_bytes, handshakeMsgEndPos)
+
+	// Seek past some headers.
+	buf, err = seek(buf, handshakeHeaderLength_bytes+clientVersionLength_bytes+clientRandomLength_bytes)
```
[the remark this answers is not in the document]
liujed: In ba374f8.

liujed:
Re-requesting review for recent changes.

akinet/tls/client_parser.go
```diff
@@ -0,0 +1,273 @@
...
+	// Now at the compression methods. The first byte gives the length of this
+	// header in bytes.
+	compressionMethodsLen_bytes, err := reader.ReadByte()
+	if err != nil {
+		return nil, 0, err
+	}
+	_, err = reader.Seek(int64(compressionMethodsLen_bytes), io.SeekCurrent)
+	if err != nil {
+		return nil, 0, err
+	}
```
mgritter, line 108: Probably a helper method would have saved some copy/paste here?

mgritter (approved):
💯 Looks good!

I like how we're better encapsulated in MemView now; further improvements to Seek orthe Reader or the Int-parsing methods won't be visible.

I'm a little dismayed that there's no way to get from a Reader to a string without some intermediary. Using StringBuilder only works if you want to read until the end of the Reader. :(

akinet/tls/client_parser.go
liujed: In 9155756.

liujed:
> I'm a little dismayed that there's no way to get from a Reader to a string without some intermediary. Using StringBuilder only works if you want to read until the end of the Reader. :(

Abstracted this into some methods on Reader for when we come up with a better way of getting a string (9155756).

PR #104, opened 2021-12-22
Support for building models from a time span of witnesses

liujed:
Added a request type and a concurrent version of `sampled_err.Errors`.


mgritter (approved):
Looks OK.

PR #107, opened 2022-01-22
Simplify logic for identifying type conflicts

thatplguy:
The logic for deciding when to introduce a `OneOf` relied on removing parts of `Data` objects before comparing them, which was subtle and prone to introducing conflicts if the IR changes in the future.

This PR refactors the logic and adds comments.  The semantics are the same.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -200,14 +200,14 @@ func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
 			return m.meldStruct(v.Struct, srcStruct.Struct)
 		} else {
 			hasConflict = true
-			return recordConflict(dst, src)
+			return m.recordConflict(dst, src)
 		}
 	case *pb.Data_List:
 		if srcList, ok := src.Value.(*pb.Data_List); ok {
 			return m.meldList(v.List, srcList.List)
 		} else {
 			hasConflict = true
-			return recordConflict(dst, src)
+			return m.recordConflict(dst, src)
 		}
 	case *pb.Data_Optional:
 		switch opt := v.Optional.Value.(type) {
@@ -229,7 +229,7 @@ func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
 			}
 			return nil
 		default:
-			return recordConflict(dst, src)
+			return m.recordConflict(dst, src)
 		}
 	case *pb.Data_Oneof:
 		hasConflict = true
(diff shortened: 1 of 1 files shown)


spec_util/meld.go
```diff
@@ -308,50 +308,37 @@ func dataEqual(dst, src *pb.Data) bool {
...
-		assignDataFormats(src, srcDataFormats)
-		assignDataFormats(dst, dstDataFormats)
-
-		// Reinstate original example values
-		src.ExampleValues = srcExampleValues
-		dst.ExampleValues = dstExampleValues
+	cmpDst := &pb.Primitive{Value: dst.Value, FormatKind: dst.FormatKind}
+	cmpSrc := &pb.Primitive{Value: src.Value, FormatKind: src.FormatKind}
+	return proto.Equal(cmpDst, cmpSrc)
+}
```
liujed, line 321: An unformatted int should subsume a formatted int, and should therefore be compatible. Similarly for other types, of course.

spec_util/meld.go
```diff
@@ -614,3 +580,34 @@ func (m *melder) meldList(dst, src *pb.List) error {
...
+	for k := range src.Formats {
+		mergedDataFormats[k] = true
+	}
+	for k := range dst.Formats {
+		mergedDataFormats[k] = true
+	}
+	if len(mergedDataFormats) > 0 {
+		dst.Formats = mergedDataFormats
+	}
+}
```
liujed, line 613: When merging a formatted int with an unformatted int, the result should be an unformatted int. This appears to produces a formatted int. Similarly for other types, of course.

liujed (commented):
The handling of unformatted primitives doesn't look quite right.

spec_util/meld.go
thatplguy: Hm, true.  I don't think the existing code does that, though.  If you don't mind, I'll land this PR with just the refactoring and then do another to properly move up the type lattice during melding.

thatplguy: I'll add some test cases to check this behavior first though.

spec_util/meld.go
thatplguy: As it stands, primitives are only merged if the base type and format kind are equal; the cases you're talking about currently result in a type conflict.

spec_util/meld.go
thatplguy: Added tests confirming that the existing behavior does not merge up the type lattice: https://github.com/akitasoftware/superstar/pull/636

liujed (approved):

PR #108, opened 2022-01-23
Fix melding of one-ofs

liujed:
Previously, when melding one-ofs together, the variants were just lumped together. Now, we go deeper into the one-ofs by ensuring that any variants that are structs get melded together, and similarly for variants that are lists.


thatplguy (approved):

thatplguy:
Closed in favor of #109.

PR #110, opened 2022-01-25
When melding primitives, join up the type lattice

thatplguy:
Right now, melding an int32 primitive with another int32 primitive with a data
format will result in a conflict.  However, this should simply result in int32.

The same is true for other base types, and also between different kinds of base
types.  As another example, melding int32 and uint32 should result in int64,
not a conflict.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -279,9 +279,7 @@ func haveCompatibleTypes(dst, src *pb.Primitive) bool {
 
 	// Types are compatible if the base types can join and the format kinds
 	// are equal.
-	cmpDst := &pb.Primitive{Value: baseJoin.Value, FormatKind: dst.FormatKind}
-	cmpSrc := &pb.Primitive{Value: baseJoin.Value, FormatKind: src.FormatKind}
-	if proto.Equal(cmpDst, cmpSrc) {
+	if dst.FormatKind == src.FormatKind {
 		return true
 	}
 
@@ -305,6 +303,9 @@ func joinBaseTypes(dst, src *pb.Primitive) *pb.Primitive {
 		return &pb.Primitive{Value: dst.Value}
 	}
 
+	// NOTE(cns): When the CLI builds witnesses from wire traffic, it parses integers
+	// as int64 whenever possible and only falls back to uint64 for values >= 2^63.
+	// However, we could see other behavior from uploaded specs.
 	switch dst.Value.(type) {
 	case *pb.Primitive_Int32Value:
 		switch src.Value.(type) {
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -264,8 +264,8 @@ func (m *melder) meldAndRehashOption(oneof *pb.OneOf, oldHash string, option *pb
 	return nil
 }
 
-// Two prims have compatible types if they have the same base type (in their
-// Value field) and the same data format kind, if any.
+// Two prims have compatible types if their base types join (in their
+// Value field) and they have the same data format kind, if any.
 func haveCompatibleTypes(dst, src *pb.Primitive) bool {
 	if dst == nil || src == nil {
 		return false
@@ -311,11 +311,17 @@ func joinBaseTypes(dst, src *pb.Primitive) *pb.Primitive {
 		switch src.Value.(type) {
 		case *pb.Primitive_Int64Value, *pb.Primitive_Uint32Value:
 			return &pb.Primitive{Value: &pb.Primitive_Int64Value{Int64Value: &pb.Int64{}}}
+		case *pb.Primitive_FloatValue:
+			return &pb.Primitive{Value: &pb.Primitive_FloatValue{FloatValue: &pb.Float{}}}
+		case *pb.Primitive_DoubleValue:
+			return &pb.Primitive{Value: &pb.Primitive_DoubleValue{DoubleValue: &pb.Double{}}}
 		}
 	case *pb.Primitive_Int64Value:
 		switch src.Value.(type) {
 		case *pb.Primitive_Int32Value, *pb.Primitive_Uint32Value:
 			return &pb.Primitive{Value: &pb.Primitive_Int64Value{Int64Value: &pb.Int64{}}}
+		case *pb.Primitive_FloatValue, *pb.Primitive_DoubleValue:
+			return &pb.Primitive{Value: &pb.Primitive_DoubleValue{DoubleValue: &pb.Double{}}}
 		}
 	case *pb.Primitive_Uint32Value:
 		switch src.Value.(type) {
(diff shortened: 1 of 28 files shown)


spec_util/meld.go
```diff
@@ -271,9 +271,76 @@ func haveCompatibleTypes(dst, src *pb.Primitive) bool {
...
+		case *pb.Primitive_Int32Value:
+			return &pb.Primitive{Value: &pb.Primitive_Int64Value{Int64Value: &pb.Int64{}}}
+		case *pb.Primitive_Uint64Value:
+			return &pb.Primitive{Value: &pb.Primitive_Uint64Value{Uint64Value: &pb.Uint64{}}}
+		}
+	case *pb.Primitive_Uint64Value:
+		switch src.Value.(type) {
+		case *pb.Primitive_Uint32Value:
+			return &pb.Primitive{Value: &pb.Primitive_Uint64Value{Uint64Value: &pb.Uint64{}}}
+		}
```
mgritter, line 330: This section looks good.

I was a bit concerned that we can't join Int64 with UInt64.  So I checked the parsing code, which for JSON numbers tries Int64 first, then Int64 again, then Uint64, then float, then bool.  (CategorizeString is called only after json.Number.Int64() fails.)

So now I'm a bit confused about how we get 32-bit integers at all?  Or don't we?  (For uploaded specs?)

It might be worth commenting above that our assumption while parsing is that everything that can fit into a signed int64 is of that type, so uint64 is really only used for values >= 2^63.

mgritter (approved):

spec_util/meld.go
```diff
@@ -271,9 +271,76 @@ func haveCompatibleTypes(dst, src *pb.Primitive) bool {
 		return false
 	}
 
-	cmpDst := &pb.Primitive{Value: dst.Value, FormatKind: dst.FormatKind}
-	cmpSrc := &pb.Primitive{Value: src.Value, FormatKind: src.FormatKind}
-	return proto.Equal(cmpDst, cmpSrc)
+	// First, check that the base types can join.
```
liujed, line 274: This changes the advertised behaviour of `haveCompatibleTypes`. Please remember to update its documentation.

spec_util/meld.go
```diff
@@ -271,9 +271,76 @@ func haveCompatibleTypes(dst, src *pb.Primitive) bool {
...
+		return true
+	}
+
+	return false
+}
+
+// Returns a new Primitive with the Value set as the type-theoretic join of
+// dst.Value and src.Value.  For example, join(int32, uint32) = int64.
+// Returns nil if no such join exists, e.g. join(int64, uint64) = nil.
+func joinBaseTypes(dst, src *pb.Primitive) *pb.Primitive {
```
liujed, line 300: Do we want to also join ints with floating points? Maybe not if the int has a format kind attached, but it may make sense otherwise.

If we choose not to join ints with floating points, the reasoning should at least be documented.

spec_util/meld.go
```diff
@@ -510,7 +580,7 @@ func (m *melder) meldList(dst, src *pb.List) error {
 // Assumes dst.value == src.value.
 // Meld data formats, tracking data, etc. from src to dst.
 // XXX(cns): In some cases, this modifies src as well as dst :/
-func (m *melder) meldPrimitive(dst, src *pb.Primitive) {
+func (m *melder) meldPrimitive(dst, src *pb.Primitive) error {
```
liujed, line 583: When can we expect this to (not) return an error? When `haveCompatibleTypes(dstPrim, srcPrim)`? Comment should say. (Or better—but less scrappy—roll `haveCompatibleTypes` into this method somehow.)

spec_util/meld.go
```diff
@@ -292,7 +359,10 @@ func (m *melder) recordConflict(dst, src *pb.Data) error {
 
 	if arePrims && haveCompatibleTypes(dstPrim, srcPrim) {
 		// No conflict.  Merge primitive metadata.
-		m.meldPrimitive(dstPrim, srcPrim)
+		err := m.meldPrimitive(dstPrim, srcPrim)
+		if err != nil {
+			return err
+		}
```
liujed, line 365: The callers of `recordConflict` all set `hasConflict = true` right before the call. But if we are able to merge primitves, then there is no actual conflict. Seems like we need to either push the merging of primitives into the callers, or return a signal from `recordConflict` that no conflict actually occurred.

liujed:
Working on a review. I think some changes need to be made. Please don't merge yet.

liujed (changes requested):
Looks like we need to account for the fact that when `recordConflict` is called, a conflict no longer necessarily happens.

Additionally, I think this PR should address this TODO in `meldOneOfVariant`: https://github.com/akitasoftware/akita-libs/blob/85dbbc9c5fe1164f9b6f1265ef21adae837ce452/spec_util/meld.go#L599-L602

spec_util/meld.go
thatplguy: Sounds good, I added the comment.

What was your concern about not being able to join Int64 and UInt64?  That's what I had expected -- to join, we'd need a signed integer type large enough to represent all values in Int64 and UInt64, which doesn't exist in our (admittedly limited) type system.

spec_util/meld.go
```diff
@@ -271,9 +271,76 @@ func haveCompatibleTypes(dst, src *pb.Primitive) bool {
...
+	cmpSrc := &pb.Primitive{Value: baseJoin.Value, FormatKind: src.FormatKind}
+	if proto.Equal(cmpDst, cmpSrc) {
+		return true
+	}
+
+	// Types are compatible if the base types can join and least one type does
+	// not have any data formats identified.
+	if dst.FormatKind == "" || src.FormatKind == "" {
+		return true
+	}
```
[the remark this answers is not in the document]
thatplguy: Doh!  I can get rid of proto.Equal entirely.  It boils down to comparing the format kinds, but I forgot to update that part.

spec_util/meld.go
mgritter: I wasn't sure when we decided something was an int64 vs. an uint64.  The answer appears to be that we always assume it's signed, unless it's too large for that.  But not being able to merge -5000000000 with +5000000000 because they are a signed and unsigned int64 would have been a problem.

spec_util/meld.go
thatplguy: I had been thinking about data formats, but now I'm not sure there's a meaningful difference between discarding formats within the same base type vs. from in to float.  I'll extend the join.

spec_util/meld.go
thatplguy: Even before this change, `recordConflict` would merge primitives in some cases without introducing a `oneof`.  Looks like the existing code uses `hasConflict` to avoid merging examples on conflicts, but `recordConflict` takes care of merging examples when there isn't a conflict.

I'm not going to refactor that as part of this PR.

spec_util/meld.go
thatplguy: I considered it but didn't want to do more refactoring.  `meldPrimitive` is called exactly once and isn't exposed outside this package, so I'm not too worried.  I updated the comment.

liujed (commented):
This TODO isn't addressed yet: https://github.com/akitasoftware/akita-libs/blob/85dbbc9c5fe1164f9b6f1265ef21adae837ce452/spec_util/meld.go#L599-L602

(GitHub won't let me point to it like in the other comments because the PR doesn't touch that part of the code.)

thatplguy:
Thanks, forgot about that.  TODO is fixed in 2e2c30c.

spec_util/meld_test.go
```diff
@@ -485,3 +505,85 @@ func TestMeldDataFormats(t *testing.T) {
...
+		left := tc.left
+		right := tc.right
+		err := MeldData(left, right)
+		assert.NoError(t, err, "[%s] failed to meld", tc.name)
+
+		if diff := cmp.Diff(tc.expected, left, cmpWitnessOptions...); diff != "" {
+			t.Errorf("[%s] found diff:\n %v", tc.name, diff)
+		}
+
+	}
```
liujed, line 588: I've seen meld tests, like the one below, that suggest that meld should be commutative. Why not test both directions here instead of having two copies of the test?

https://github.com/akitasoftware/akita-libs/blob/85dbbc9c5fe1164f9b6f1265ef21adae837ce452/spec_util/meld_test.go#L239-L270

spec_util/meld_test.go
```diff
@@ -291,7 +291,7 @@ func witnessLess(w1, w2 *pb.Witness) bool {
 	return proto.MarshalTextString(w1) < proto.MarshalTextString(w2)
 }
 
-func TestMeldDataFormats(t *testing.T) {
+func TestMeldPrimitives(t *testing.T) {
```
liujed, line 294: Similarly for this test.

liujed (approved):
Small nit regarding tests, but looks good.

spec_util/meld_test.go
thatplguy: Good idea, fixed in 3089a2d.

PR #111, opened 2022-01-30
Add support for contextless visitors

liujed:
Allows us to avoid the memory cost of maintaining contexts when they're not needed.

### DIFF
--- a/visitors/http_rest/spec_visitor_context.go
+++ b/visitors/http_rest/spec_visitor_context.go
@@ -540,3 +540,103 @@ func (c stackVisitorContext) setHttpAuthType(at pb.HTTPAuth_HTTPAuthType) {
 func (c stackVisitorContext) setTopLevelDataIndex(i int) {
 	c.Delegate().setTopLevelDataIndex(i)
 }
+
+type DummyVisitorContext struct{}
+
+var _ SpecVisitorContext = (*DummyVisitorContext)(nil)
+
+func NewDummyVisitorContext() *DummyVisitorContext {
+	return &DummyVisitorContext{}
+}
+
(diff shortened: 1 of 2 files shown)


visitors/http_rest/spec_visitor_context.go
```diff
@@ -540,3 +540,103 @@ func (c stackVisitorContext) setHttpAuthType(at pb.HTTPAuth_HTTPAuthType) {
...
+
+func (c *DummyVisitorContext) EnterArray(arrayNode interface{}, elementIndex int) visitors.Context {
+	return c
+}
+
+func (c *DummyVisitorContext) EnterMapValue(mapNode, mapKey interface{}) visitors.Context {
+	return c
+}
+
+func (*DummyVisitorContext) GetPath() visitors.ContextPath {
```
thatplguy, line 564: Are any of these methods called only by the user?  Would it be worth calling `panic` rather than returning a value for those?

It would be nice to catch the case where someone modifies an existing visitor method and accidentally uses the dummy context without changing `NewContext()`.

thatplguy (approved):

visitors/http_rest/spec_visitor_context.go
liujed: I considered this, but a bunch of methods are called internally by the visitor framework, and I think it is better to have them all uniformly return dummy values, rather than have a select few calling `panic`. This said, I think the visitor framework can have a special case to avoid these calls when the dummy context is used. This would allow all of these methods panic if called.

thatplguy: Makes sense.  Want to put a TODO in the code (or make a ticket) to remind us to do this in the future?

liujed: In 6d7fca1.

PR #113, opened 2022-02-04
Improve path parameter annotations

thatplguy:
Path parameter annotations used to include special symbols `{arg}`, `*`
(wildcard), and `^` (prevent path parameter at this position), and they used to
match by longest prefix.

This PR makes two syntax changes:
 - removes `^` (it was never used in practice)
 - adds `**` (like `*`, but matches zero or more path segments)

It also makes several changes to the matching semantics:
 - annotations are now exact match rather than prefix match (prefix match is
   recoverable using `**`)
 - matches ignore trailing slashes

### DIFF
--- a/path_pattern/components.go
+++ b/path_pattern/components.go
@@ -1,7 +1,7 @@
 package path_pattern
 
 import (
-	"strings"
+	"regexp"
 )
 
 // Represents a path component value, which can be either a concrete string Val
@@ -23,17 +23,7 @@ func (v Val) String() string {
 }
 
 func (v Val) Regexp() string {
-	return "(" + escape(v.String()) + ")"
-}
-
-// Escapes special regexp characters in s.
-func escape(s string) string {
-	specialChars := []string{"\\",".","+","*","?","(",")","|","[","]","{","}","^","$"}
-	rv := s
-	for _, c := range specialChars {
-		rv = strings.ReplaceAll(rv, c, "\\" + c)
-	}
-	return rv
+	return "(" + regexp.QuoteMeta(v.String()) + ")"
 }
 
 type Var string
--- a/path_pattern/components.go
+++ b/path_pattern/components.go
@@ -1,10 +1,15 @@
 package path_pattern
 
+import (
+	"strings"
+)
+
 // Represents a path component value, which can be either a concrete string Val
 // or a Var.
 type Component interface {
 	Match(string) bool
 	String() string
+	Regexp() string
 }
 
 type Val string
@@ -17,6 +22,20 @@ func (v Val) String() string {
 	return string(v)
 }
 
+func (v Val) Regexp() string {
+	return "(" + escape(v.String()) + ")"
+}
+
+// Escapes special regexp characters in s.
+func escape(s string) string {
+	specialChars := []string{"\\",".","+","*","?","(",")","|","[","]","{","}","^","$"}
+	rv := s
+	for _, c := range specialChars {
+		rv = strings.ReplaceAll(rv, c, "\\" + c)
+	}
+	return rv
+}
+
 type Var string
 
 func (Var) Match(c string) bool {
@@ -26,6 +45,10 @@ func (Var) Match(c string) bool {
 
 func (v Var) String() string {
 	return "{" + string(v) + "}"
+}
+
+func (v Var) Regexp() string {
+	return "([^/]+)"
 }
 
 // A component that matches any path argument, either a concrete value or a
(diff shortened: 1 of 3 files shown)


path_pattern/components.go
```diff
@@ -17,6 +22,20 @@ func (v Val) String() string {
...
+
+// Escapes special regexp characters in s.
+func escape(s string) string {
+	specialChars := []string{"\\",".","+","*","?","(",")","|","[","]","{","}","^","$"}
+	rv := s
+	for _, c := range specialChars {
+		rv = strings.ReplaceAll(rv, c, "\\" + c)
+	}
+	return rv
+}
```
liujed, line 37: Can we use this instead? https://pkg.go.dev/regexp?utm_source=godoc#QuoteMeta

liujed (approved):

path_pattern/pattern.go
```diff
@@ -23,33 +24,45 @@ func (p *Pattern) UnmarshalText(data []byte) error {
...
+	for _, piece := range p {
+		pieces = append(pieces, piece.Regexp())
+	}
+	return regexp.MustCompile("^" + strings.Join(pieces, "/") + "$")
+}
+
+// Match happens if the pattern exactly matches the string.
 func (p Pattern) Match(v string) bool {
-	parts := strings.Split(v, "/")
+	r := p.regexp()
```
mgritter, line 37: I think we saw on the CLI that recompiling the regexp each time can be a lot of load; can we amortize it across multiple path matches?

path_pattern/pattern.go
```diff
@@ -23,33 +24,45 @@ func (p *Pattern) UnmarshalText(data []byte) error {
...
 	}
-	return true
+	return s
 }
 
 // Converts a string pattern "/v1/{arg2}" to Pattern.
 func Parse(v string) Pattern {
-	parts := strings.Split(v, "/")
+	parts := strings.Split(removeTrailingSlashes(v), "/")
 	result := make(Pattern, 0, len(parts))
```
mgritter, line 59: I am not really clear on how trailing slashes are handled.  From the test it looks like
/foo/{arg2}
matches
/foo/a/
which is fine, but maybe a comment somewhere should set expectations.
?

mgritter (approved):
Looks good. Should try to avoid recompiling the regexp on each compare, if we can.

This doesn't seem to be part of the PR?  Or if it is, it's too subtle for me to understand.
> a matching annotation with concrete path components prevents path parameter
inference from generalizing those components.

path_pattern/pattern.go
```diff
@@ -23,33 +24,45 @@ func (p *Pattern) UnmarshalText(data []byte) error {
...
-	parts := strings.Split(v, "/")
+	r := p.regexp()
+	return r.MatchString(removeTrailingSlashes(v))
+}
 
-	compLen := len(p)
-	if compLen > len(parts) {
-		return false
-	}
+func (p Pattern) MatchWithGroup(v string) (bool, []string) {
```
liujed, line 41: A comment documenting the returned `[]string` value (ideally with examples) would be useful.

path_pattern/components.go
```diff
@@ -1,10 +1,15 @@
...
+import (
+	"regexp"
+)
+
 // Represents a path component value, which can be either a concrete string Val
 // or a Var.
 type Component interface {
 	Match(string) bool
 	String() string
+	Regexp() string
```
liujed, line 12:
```
// Returns a parenthesized regular expression for this component.
```

thatplguy:
> This doesn't seem to be part of the PR? Or if it is, it's too subtle for me to understand.
> 
> > a matching annotation with concrete path components prevents path parameter
> > inference from generalizing those components.

That's part of a companion PR -- the syntax of patterns and semantics of matching lives in `akita-libs`, but the interpretation of what to do with a matched pattern doesn't.  I'll change the description to be more clear.

path_pattern/pattern.go
thatplguy: It matches ignoring trailing slashes, so `/v1/{arg}` and `/v1/{arg}/` would both match `/v1/foo` and `/v1/foo/`.  I added a comment.

path_pattern/pattern.go
thatplguy: Good idea, fixed in 591c5e0.

path_pattern/pattern.go
thatplguy: Fixed in 591c5e0.

path_pattern/components.go
thatplguy: Added in 591c5e0.

PR #126, opened 2022-04-18
Support tag multisets for tagging specs with versions

thatplguy:
This PR extends tags to support multiple values per tag.  APIs will still
reject requests with tag multisets, but responses may now contain tag
multisets.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -255,7 +255,7 @@ type GetSpecMetadataResponse struct {
 
 	State APISpecState `json:"state"`
 
-	Tags map[tags.Key]string `json:"tags"`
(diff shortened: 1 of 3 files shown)


api_schema/schema.go
```diff
@@ -255,7 +255,7 @@ type GetSpecMetadataResponse struct {
 
 	State APISpecState `json:"state"`
 
-	Tags map[tags.Key]string `json:"tags"`
+	Tags map[tags.Key][]string `json:"tags"`
```
liujed, line 258: Feels like we ought to introduce a type abstraction for this.

liujed (approved):

PR #127, opened 2022-04-20
Make tag multisets backwards compatible

thatplguy:
After thinking about it, I'm not convinced it was safe to change the tag fields
in 10b6ba80886d956ba3d08a31a8b63bd99555b92c.

This PR reverts those changes, introduces new tags abstractions that
distinguishes between singleton tag maps and multiset tag maps, and adds new
response fields where we expect to expose multiset tag maps.

### DIFF
--- a/tags/singleton_tags.go
+++ b/tags/singleton_tags.go
@@ -1,0 +1,38 @@
+package tags
+
+import "github.com/pkg/errors"
+
+// SingletonTags maps tags to single values.
+type SingletonTags map[Key]string
+
+func (ts SingletonTags) AsTags() Tags {
+	tags := make(Tags, len(ts))
+	for t, v := range ts {
+		tags.SetSingleton(t, v)
+	}
+	return tags
+}
+
+// FromPairsSingleton returns a map from parsing a list of "key=value" pairs.
+// Produces an error if any element of the list is improperly formatted,
+// or if any key is given more than once.
+// The caller must emit an appropriate warning if any keys are reserved.
+func FromPairs(pairs []string) (SingletonTags, error) {
+	tags, err := FromPairsMultivalue(pairs)
+	if err != nil {
+		return nil, err
+	}
+
+	results := make(SingletonTags, len(tags))
+	for k, vs := range tags {
+		vSlice := vs.AsSlice()
+		if len(vSlice) == 0 {
+			continue
+		} else if len(vSlice) > 1 {
+			return nil, errors.Errorf("tag with key %s specified more than once", k)
+		}
+		results[k] = vSlice[0]
+	}
+
+	return results, nil
+}
--- a/tags/singleton_tags.go
+++ b/tags/singleton_tags.go
@@ -13,7 +13,7 @@ func (ts SingletonTags) AsTags() Tags {
 	return tags
 }
 
-// FromPairsSingleton returns a map from parsing a list of "key=value" pairs.
+// Returns a SingletonTags from parsing a list of "key=value" pairs.
 // Produces an error if any element of the list is improperly formatted,
 // or if any key is given more than once.
 // The caller must emit an appropriate warning if any keys are reserved.
--- a/tags/tags.go
+++ b/tags/tags.go
@@ -8,19 +8,20 @@ import (
 )
 
 type Key = tags.Key
-type Values = []string
+type Tags map[Key][]string
+type SingletonTags map[Key]string
 
-// FromPairs returns a map from parsing a list of "key=value" pairs.
+// Returns a map from parsing a list of "key=value" pairs.
 // Produces an error if any element of the list is improperly formatted,
 // or if any key is given more than once.
 // The caller must emit an appropriate warning if any keys are reserved.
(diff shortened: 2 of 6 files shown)


liujed:
(10b6ba8 refers to #126.)

liujed (approved):

thatplguy:
Requesting a second review after adding utility methods for the new tag abstractions.

tags/singleton_tags.go
```diff
@@ -0,0 +1,38 @@
...
+		tags.SetSingleton(t, v)
+	}
+	return tags
+}
+
+// FromPairsSingleton returns a map from parsing a list of "key=value" pairs.
+// Produces an error if any element of the list is improperly formatted,
+// or if any key is given more than once.
+// The caller must emit an appropriate warning if any keys are reserved.
+func FromPairs(pairs []string) (SingletonTags, error) {
```
liujed, line 20: Function name doesn't match comment.

tags/singleton_tags.go
```diff
@@ -0,0 +1,38 @@
...
+func (ts SingletonTags) AsTags() Tags {
+	tags := make(Tags, len(ts))
+	for t, v := range ts {
+		tags.SetSingleton(t, v)
+	}
+	return tags
+}
+
+// FromPairsSingleton returns a map from parsing a list of "key=value" pairs.
+// Produces an error if any element of the list is improperly formatted,
```
liujed, line 17:
```suggestion
// FromPairsSingleton returns a SingletonTags from parsing a list of "key=value"
// pairs. Produces an error if any element of the list is improperly formatted,
```

tags/tags.go
```diff
@@ -8,44 +8,102 @@ import (
...
+type Tags map[Key]ValueSet
+
+// Sets the given key to the given values.  Values are copied.
+func (t Tags) Set(key tags.Key, values []string) {
+	t[key] = NewValueSet(values...)
+}
+
+// Sets the given key to the given value.
+func (t Tags) SetSingleton(key tags.Key, value string) {
+	t.Set(key, []Value{value})
```
liujed, line 23:
```suggestion
	t.Set(key, NewValueSet(value))
```

tags/tags.go
```diff
@@ -8,44 +8,102 @@ import (
...
+		t.Set(key, values.AsSlice())
 	}
+}
 
-	results := make(map[Key]string, len(multiset))
-	for k, vs := range multiset {
-		if len(vs) > 1 {
-			return nil, errors.Errorf("tag with key %s specified more than once", k)
+// Removes any tag from t that doesn't exist in t2.  For any key k in both t
+// and t2, t[k] is remapped to the intersection of their values.
```
liujed, line 40:
```suggestion
// Removes any tag from t that doesn't exist in t2.  For any key k in both t
// and t2, t[k] is remapped to the intersection of their values. If the
// intersection is empty, then k is removed.
```

tags/tags.go
```diff
@@ -8,44 +8,102 @@ import (
...
+	return rv
 }
 
-// FromPairsMultiset returns a map from parsing a list of "key=value" pairs.
+// FromPairs returns a map from parsing a list of "key=value" pairs.
 // Produces an error if any element of the list is improperly formatted.
 // The caller must emit an appropriate warning if any keys are reserved.
-func FromPairsMultiset(pairs []string) (map[Key]Values, error) {
-	results := make(map[Key]Values, len(pairs))
+func FromPairsMultivalue(pairs []string) (Tags, error) {
```
liujed, line 97: Function name doesn't match comment.

tags/tags.go
```diff
@@ -8,44 +8,102 @@ import (
...
+	for tag, values := range t {
+		if v, exists := values.GetFirst(); exists {
+			rv[tag] = v
+		}
+	}
+	return rv
 }
 
-// FromPairsMultiset returns a map from parsing a list of "key=value" pairs.
+// FromPairs returns a map from parsing a list of "key=value" pairs.
```
liujed, line 94:
```suggestion
// FromPairs returns a Tags from parsing a list of "key=value" pairs.
```

tags/value_sets.go
```diff
@@ -0,0 +1,88 @@
...
+	vs[v] = struct{}{}
+}
+
+// Adds all values from other to v.
+func (vs ValueSet) AddAll(other ValueSet) {
+	for v, _ := range other {
+		vs.Add(v)
+	}
+}
+
```
liujed, line 46: Isn't this just `Union`? (This is called from `Tags.Union` and from the unit tests for `ValueSet`. Suggest removing the affected unit tests and changing the other call to use `Union`.)
```suggestion
```

liujed (approved):
Looks good. Biggest suggested change is to remove `ValueSet.AddAll`. Otherwise, just a few small comments, mostly to help respect type abstractions and to maintain consistency between documentation comments and function names.

tags/singleton_tags.go
thatplguy: I immediately regret trying this style.

PR #132, opened 2022-05-19
Added `String()` for Akita IDs

liujed:
Avoids having to call `akid.String` everywhere.

Also fixed some warnings in `akid_test.go`.

### DIFF
--- a/akid/akids.go
+++ b/akid/akids.go
@@ -121,6 +121,10 @@ func (APISpecID) GetType() string {
 	return APISpecTag
 }
 
+func (id APISpecID) String() string {
+	return String(id)
+}
+
 func NewAPISpecID(ID uuid.UUID) APISpecID {
 	return APISpecID{baseID(ID)}
 }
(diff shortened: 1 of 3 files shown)


akid/akids.go
```diff
@@ -121,6 +121,10 @@ func (APISpecID) GetType() string {
 	return APISpecTag
 }
 
+func (id APISpecID) String() string {
+	return String(id)
```
mgritter, line 125: Seeing the unqualified function name everywhere was super-confusing.  :)  Nothing to fix here, though.

mgritter (approved):
Looks good.

I thought about it for a bit and I don't see a Go solution for having the base class call a derived-type method (other than passing the interface as a separate argument, which is a nonstarter here.)

Could be resolved by having the akid "type" be a member of the base class instead of retrieved via a function, but that would be a lot of churn too.

akid/akids.go
liujed: > Could be resolved by having the akid "type" be a member of the base class instead of retrieved via a function, but that would be a lot of churn too.

Yeah, this occurred to me too, but I decided against it—there's something satisfying about having AKIDs be represented by just UUIDs.

liujed:
> Could be resolved by having the akid "type" be a member of the base class instead of retrieved via a function, but that would be a lot of churn too.

Yeah, this occurred to me too, but I decided against it—there's something satisfying about having AKIDs be represented by just UUIDs.

PR #135, opened 2022-05-27
Represent optionality in field path

liujed:
Also tweaked string representation of keys and values of inferred maps


mgritter (approved):
Looks good.

PR #137, opened 2022-06-07
Tweaked one-of diffing

liujed:
Lifted additions and removals at the level of one-of variants to changes at the level of the enclosing one-of.

### DIFF
--- a/visitors/http_rest_diff/spec_diff_visitor.go
+++ b/visitors/http_rest_diff/spec_diff_visitor.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/akitasoftware/akita-libs/visitors"
 	"github.com/akitasoftware/akita-libs/visitors/go_ast_pair"
 	"github.com/akitasoftware/akita-libs/visitors/http_rest"
+	"golang.org/x/exp/maps"
 )
 
 // A SpecPairVisitor with hooks for processing each difference found between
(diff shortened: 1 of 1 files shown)


visitors/http_rest_diff/spec_diff_visitor.go
```diff
@@ -210,9 +213,39 @@ func (*DefaultSpecDiffVisitorImpl) EnterOneOfs(self interface{}, ctx http_rest.S
...
+			if IsSameData(leftOption, rightOption) {
+				delete(rightOptions, rightKey)
+				continue OUTER
+			}
+		}
+
+		// No match found for leftOption.
+		return true
+	}
+	return false
```
mgritter, line 246: I don't think this is right.

Consider leftOption = {A} and rightOption = {A, B}

I think you meant 
```
return len( rightOption ) == 0
```

liujed: Isn't this covered by the length check above the loop?

mgritter: Hm, OK, you're right, each loop should decrease the length of rightOptions by one or terminate.

mgritter (approved):

visitors/http_rest_diff/spec_diff_visitor.go
liujed: Let's port to F* real quick and verify this! 😄

mgritter:
OK, is the problem that the visitor previously visited each option in an unpredictable order, leading to false positives, but now it removes all those false positives -- but, the add and remove will still look the same for oneOfs that actually have differences?

liujed:
Answered here: https://akitasoftware.slack.com/archives/CSL17U42G/p1654565899003119.

PR #138, opened 2022-06-08
Extend map heuristics

thatplguy:
We use heuristics to decide when a JSON object in an API is a map
(dictionary) with variable data keys rather than a fixed part of the
API.

This PR extends the heuristics to account for objects with fields that
begin with numbers -- this covers common data formats like numbers,
datetimes, etc.  If every field in a struct beings with a number and
there are sufficient fields, we consider it to be a map.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -2,12 +2,11 @@ package spec_util
 
 import (
 	"fmt"
-	"reflect"
-	"sort"
-	"strconv"
-
 	"github.com/pkg/errors"
 	"google.golang.org/protobuf/proto"
+	"reflect"
+	"regexp"
+	"sort"
 
 	pb "github.com/akitasoftware/akita-ir/go/api_spec"
 	"github.com/akitasoftware/akita-libs/spec_util/ir_hash"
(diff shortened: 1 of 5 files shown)


spec_util/meld.go
```diff
@@ -478,18 +480,42 @@ func structShouldBeMap(struc *pb.Struct) bool {
...
+	if len(s) == 0 {
+		return false
+	}
+
+	if _, err := strconv.Atoi(s[0:1]); err != nil {
+		return false
+	}
+
+	return true
+}
```
liujed, line 517: Why not use a regular expression `^\d`? Alternatively, could compare `'0' <= s[0] && s[0] <= '9'` instead of calling `Atoi`.

spec_util/meld.go
```diff
@@ -478,18 +480,42 @@ func structShouldBeMap(struc *pb.Struct) bool {
...
 		}
+		if !startsWithNumber(fieldName) {
+			allFieldsStartWithNumbers = false
+		}
+	}
+
+	// A struct should be a map if all its fields start with numbers and there
+	// are a sufficient number of fields.  Because many programming languages
+	// disallow struct names starting with numbers, the number of fields is
+	// lower.
```
liujed, line 499: An alternative might be to convert to a map if there are enough field names starting with a number, even in the presence of field names starting with other characters.

Another heuristic might be to look at how many fields have certain kinds of punctuation in their names.

What's here seems fine, though.

liujed (approved):
Looks good

PR #145, opened 2022-07-18
Fix capitalization 😞

liujed:


liujed:
If only there were a way to use `QualifiedServiceID`s as `ServiceID`s... 😞

mgritter (approved):

PR #147, opened 2022-07-27
Add akid for alert monitors

thatplguy:


mgritter (approved):
Don't forget to change the Python side too, please.

PR #148, opened 2022-07-29
Print context when we have nil data during summarization

liujed:


thatplguy (approved):

spec_summary/summarize.go
```diff
@@ -123,6 +126,11 @@ func (v *specSummaryVisitor) LeaveData(self interface{}, context vis.SpecVisitor
 	} else if meta := spec_util.HTTPCookieFromData(d); meta != nil {
 		v.methodSummary.DirectedFilters.Increment(direction, ParamFilter, meta.Key)
 	} else {
+		if d == nil {
+			glog.Errorf("Context of nil data: %v", strings.Join(slices.Map(context.GetPath(), func(e ContextPathElement) string {
+				return fmt.Sprintf("[%s] %s", reflect.TypeOf(e.AncestorNode), e.OutEdge)
```
mgritter, line 131: I would prefer that we alert on this error (we could even give it a more distinctive tag to make it easy!) rather than continuing to fail to build the model.

liujed: In ab6f4ac and 4ca702e.

PR #150, opened 2022-08-20
Memory-usage improvements

liujed:
Two improvements to memory usage:
* Don't buffer HTTP requests and responses during parsing. Previously, we were buffering the entire request/response so that it could be emitted as raw data when parsing failed. Instead we now just emit a count of how much data was consumed.
* Read HTTP payloads from a fixed-size buffer pool.

Also added some documentation to the HTTP and TCP parsers, including for a bug in the HTTP response parser.


buffer_pool/buffer.go
```diff
@@ -0,0 +1,246 @@
...
+	//   - all elements have length and capacity pool.chunkSize_bytes.
+	chunks [][]byte
+
+	// Contents of the buffer start at chunks[0][readOffset] (inclusive). This is
+	// where to start reading from.
+	//
+	// Invariants:
+	//   - readOffset == 0 when len(chunks) == 0.
+	//   - readOffset < pool.chunkSize_bytes when len(chunks) > 0.
+	//   - readOffset < writeOffset when len(chunks) == 1.
```
mgritter, line 63: I think you should express these in code (although I see there is no unit test in which to call it.)

While testing in our own infrastructure we could enable this invariant-checking as an assertion at the exit from ReadFrom and Write.

buffer_pool/buffer_pool.go
```diff
@@ -0,0 +1,73 @@
...
+	case result := <-pool.chunks:
+		return result
+	default:
+		return nil
+	}
+}
+
+// Releases the given chunks back to the pool.
+func (pool bufferPool) release(chunks [][]byte) {
+	// Avoid blocking, in case we somehow end up releasing more chunks than were
```
mgritter, line 63: I kind of want to zero these out (or some other sentinel value) to aid in debugging?

mgritter (approved):
Looks good. I spent some time on the buffer implementation and it looks solid, but as usual I'd like to see unit tests, particularly property-based tests, to expose any boundary conditions we're both overlooking.

akinet/tcp_parser.go
```diff
@@ -84,17 +92,17 @@ type TCPParserFactorySelector []TCPParserFactory
...
+//   - no factory is suitable
+//   - f=nil, decision=NeedMoreData, discardFront>=0
+//   - no factory returned Accept
+//   - at least one factory requested more data
+//   - discardFront is the MIN of all the discardFronts returned by factories
+//     requesting more data
+//   - f!=nil, decision=Accept, discardFront>=0
+//   - a factory has been selected
+//   - caller must discard discardFront number of bytes from input before
+//     feeding input to the parser generated by the factory.
```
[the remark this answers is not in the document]
liujed: Yeah, auto-formatting. Seems to be a new behaviour. Fixed in 191429c.

buffer_pool/buffer_pool.go
liujed: Added to `getChunk` in 8f36be6. Spent some time looking for a `memset` equivalent, but it turns out the Go compiler has a peephole optimization for zeroing out a slice: https://github.com/golang/go/commit/f03c9202c43e0abb130669852082117ca50aa9b1.

buffer_pool/buffer.go
liujed: Added in cf20c0a. Unit tests coming soon!

liujed: Unit tests in 1caff2e (refactored in 5148008).

liujed:
Requesting review of unit tests.

buffer_pool/buffer.go
liujed: Invariant-checking enabled during dogfooding in akitasoftware/akita-cli@b2dc2550cffe3022c7bc035b49aed3aa337322a2 as part of akitasoftware/akita-cli#147.

PR #152, opened 2022-09-15
Add Segment Analytics Client

versilis:
This would add a wrapper client for sending Segment events. It adds additional logging capability with glog and the ability to also send events to Mixpanel.


analytics/event.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// The timestamp of the event. Set to nil to use the current time.
+	timestamp time.Time
+
+	// Custom properties of the event.
+	properties map[string]any
+}
+
+// Returns a new event with the given name, properties, and timestamp. Given no timestamp is provided, the current time is used.
+func NewEvent(name string, properties map[string]any, timestamp ...time.Time) *Event {
```
mgritter, line 18: I'm not sure what having a separate Event object gives us? Although both Mixpanel and Segment do it, so I guess there must be some good reason.  :)

The expected use case is probably
```
analyticsClient.Track( user.Email, analytics.NewEvent(
    "my event",
    map[string]any{ 
       "cause" : "something",
    },
) )
```

so I wonder if it's better to combine the two in the common case:
```
analyticsClient.Track( user.Email, "my event",
    map[string]any{ 
       "cause" : "something",
    },
)
```

and have a separate method that can take a fully-populated Event.

mgritter (approved):
Looks good, please merge it.

analytics/client.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+type Client interface {
+	// Sends the given tracking event to Segment and Mixpanel (if enabled).
+	Track(distinctID string, event *Event) error
+}
+
+type clientImpl struct {
+	config        Config
+	segmentClient analytics.Client
```
liujed, line 18: I'd import the Segment package as `segment` to avoid confusing `analytics.Client` with the `Client` interface defined above.

analytics/event.go
```diff
@@ -0,0 +1,31 @@
...
+
+import "time"
+
+// Holds the name and properties of an analytics event.
+type Event struct {
+	// The name of the event.
+	name string
+
+	// The timestamp of the event. Set to nil to use the current time.
+	timestamp time.Time
```
liujed, line 11: Do you mean "set to 0"? `nil` is not a valid value for `time.Time`.

analytics/event.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Custom properties of the event.
+	properties map[string]any
+}
+
+// Returns a new event with the given name, properties, and timestamp. Given no timestamp is provided, the current time is used.
```
liujed, line 17:
```suggestion
// Returns a new event with the given name, properties, and timestamp. If no timestamp is provided, the current time is used.
```

analytics/event.go
liujed, line 18: Feels like a bit of an abuse of a variadic function. I'm uncomfortable that if more than one timestamp is given, those extra arguments are silently discarded. Why not just have the caller provide `time.Now()` explicitly? Or, if that's the common case, maybe have a `NewEvent(string, map[string]any)` and a `NewEventWithTime(string, map[string]any, time.Time)`. Or, alternatively, declare a `NewEvent(string, map[string]any)` and a `func (*Event) SetTime(time.Time) *Event` so that we can write `myEvent := NewEvent(...).SetTime(t)`.

analytics/config.go
```diff
@@ -0,0 +1,32 @@
...
+
+type Config struct {
+	// The key used to identify the Segment source to use
+	WriteKey string `yaml:"segment_write_key"`
+
+	// The endpoint to which the segment client connects to send events
+	SegmentEndpoint string `yaml:"segment_endpoint"`
+
+	// Data pertaining to the application such as name, version, and build
+	// If set, the specified values will globally be added to each event context
```
liujed, line 13:
```suggestion
	// If set, the specified values will be added globally to each event context
```

liujed (commented):
Feedback so far. Working on `client.go`.

analytics/event.go
versilis: I think it might be helpful for ensuring consistent event types and properties using helper functions. This might be helpful if we were to send the same event in multiple places.

So something like:
```
func NewFooEvent(value string) *Event {
	return NewEvent("foo", map[string]any{
		"foo": value,
	}),
}
```

However, I do agree that it would be nice to have a shorthand for this. I've added a new method to handle both cases in [064e35e](https://github.com/akitasoftware/akita-libs/pull/152/commits/064e35e445b2ee0ac173764b98b99415022e1fca)

analytics/client.go
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	return &clientImpl{
+		config:         config,
+		segmentClient:  segmentClient,
+		mixpanelClient: mixpanelClient,
+	}, nil
+}
+
+func (c clientImpl) TrackEvent(distinctID string, event *Event) error {
```
liujed, line 57: Please document: what is `distinctID`? Looks like this identifies the user. (Why isn't it part of `Event`?)

liujed (approved):

analytics/client.go
versilis: Updated in [f98a5fb](https://github.com/akitasoftware/akita-libs/pull/152/commits/f98a5fb1c9d9ad9fc3ebe457f7c38611db39b2b8)

analytics/event.go
versilis: Oops, I never removed this comment from an earlier revision. Fixed in [94aedbe](https://github.com/akitasoftware/akita-libs/pull/152/commits/94aedbee95b665ccf0152cdfc4b3bbb4a01a8b46)

analytics/event.go
versilis: I've updated this in [8ad54fe](https://github.com/akitasoftware/akita-libs/pull/152/commits/8ad54fe616e1834a3c31cfcd30f0fa36fe03a1e3)

analytics/client.go
versilis: This is the unique identifier for the user. I've chosen the name `distinctID` because I wanted to avoid getting it confused with our `akid.UserID`. I think it makes sense to add this to the event struct. I've updated this in [bfad548](https://github.com/akitasoftware/akita-libs/pull/152/commits/bfad548495f7e2338be10d3e5786fe25df3e4c42)

PR #160, opened 2022-10-04
Manage cookies in akinet requests/responses

thatplguy:
The akinet library has request and response structs that mirror http.Request
and http.Response.  As a convenience, the akinet structs have a Cookies field
that copies cookies from Headers.

This PR ensures that the Cookies field is populated when converting between
akinet and http objects.

### DIFF
--- a/akinet/http/parser_test.go
+++ b/akinet/http/parser_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"compress/flate"
 	"fmt"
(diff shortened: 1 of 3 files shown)


mgritter (approved):
Looks good.

What is the context where we're converting from the akinet object back to the standard library object?  Is it just for HAR files?

thatplguy:
Just HAR files, as far as I can tell.

PR #166, opened 2022-11-01
Update ListSpecsResponse and SpecInfo

liujed:
Remove unused and obsolete attributes. Added an indication of whether there are more pages of results.

### DIFF
--- a/api_schema/schema.go
+++ b/api_schema/schema.go
@@ -296,19 +296,6 @@ type SpecInfo struct {
 
 	Name string `json:"name,omitempty"`
 
-	// Whether the spec was LEARNED or LEARNED_WITH_EDITS.
-	CreationMode LearnMode `json:"creation_mode"`
-
-	// If the spec was created from a learn session, the session's ID is included.
-	// Deprecated: use learn_session_ids instead.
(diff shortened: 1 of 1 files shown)


api_schema/schema.go
```diff
@@ -296,19 +296,6 @@ type SpecInfo struct {
...
-
-	LearnSessionIDs []akid.LearnSessionID `json:"learn_session_ids,omitempty"`
-
-	// Deprecated: for old specs, tags are inherited from learn sessions.
-	// Use Tags field instead.
-	LearnSessionTags []LearnSessionTag `json:"learn_session_tags,omitempty"`
-
 	// Deprecated in favor of TagsSet, which supports multiple values
 	// per tag.
 	Tags    tags.SingletonTags `json:"tags,omitempty"`
```
mgritter, line 301: Why not remove this deprecated field too?

thatplguy (approved):

mgritter (approved):

api_schema/schema.go
liujed: The front end still uses it.

PR #168, opened 2022-11-04
Tweak FiltersToMethods

liujed:
Support mapping filter values to empty sets. Allows us to track values that exist in a model, but have no matching methods because they've all been filtered out.

### DIFF
--- a/spec_summary/filter_map.go
+++ b/spec_summary/filter_map.go
@@ -1,18 +1,23 @@
 package spec_summary
 
 import (
+	"github.com/akitasoftware/go-utils/optionals"
+	"github.com/akitasoftware/go-utils/sets"
 	. "github.com/akitasoftware/go-utils/sets"
 )
 
 // Maps all supported filter kinds to the methods that match them. The methods
 // for each filter kind are indexed by the filter's value for each method.
 //
+// A filter's value is mapped to an empty set when the value only occurs in
+// methods that have been filtered out.
+//
 // For example, if the supported filter kinds are operation and path, then the
 // method set {"GET /", "PUT /"} is represented as follows:
 //
-//	operation -> "GET" -> "GET /"
-//	             "PUT" -> "PUT /"
-//	path -> "/" -> "GET /", "PUT /"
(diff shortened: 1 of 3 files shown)


spec_summary/filter_map_test.go
```diff
@@ -12,14 +14,15 @@ func TestFiltersToMethods(t *testing.T) {
...
+	fm.InsertDirectionalFilter(RequestDirection, AuthFilter, "None", m1, ms)
+	fm.InsertDirectionalFilter(RequestDirection, AuthFilter, "None", m2, ms)
 
-	fm.filterMap.Insert(HttpMethodFilter, "GET", m1)
-	fm.filterMap.Insert(HttpMethodFilter, "POST", m2)
+	fm.filterMap.Insert(HttpMethodFilter, "GET", optionals.Some(m1))
+	fm.filterMap.Insert(HttpMethodFilter, "POST", optionals.Some(m2))
 
 	// No filters.
 	directedSummary, numMethods := fm.SummarizeWithFilters(nil)
```
mgritter, line 28: I am kind of surprised that the "with filters" case below did not result in any changes, but it was checking that the counts were equal to zero, which is what we'd expect in any case.

Is there an assert we can add to distinguish the new and old behavior?

mgritter (approved):

spec_summary/filter_map_test.go
mgritter: Hm, maybe it's because we didn't check the MethodID behavior at all?  In which case some test would be appropriate for that.

liujed: New tests in 3396096.

spec_summary/filter_map.go
```diff
@@ -177,7 +195,9 @@ func (fm FilterMap[MethodID]) Insert(filterKind FilterKind, filterValue string,
 		methodsByFilterValue[filterValue] = methods
 	}
 
-	methods.Insert(method)
```
Said182, line 180: apidump/apidump.go

PR #169, opened 2022-11-04
Tweak non-directional filters in FiltersToMethods

liujed:
Support mapping non-directional filter values to empty sets. Follows up on #168, which missed non-directional filters.

### DIFF
--- a/spec_summary/filter_map.go
+++ b/spec_summary/filter_map.go
@@ -2,7 +2,6 @@ package spec_summary
 
 import (
 	"github.com/akitasoftware/go-utils/optionals"
-	"github.com/akitasoftware/go-utils/sets"
 	. "github.com/akitasoftware/go-utils/sets"
 )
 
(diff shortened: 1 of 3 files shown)


spec_summary/filter_map.go
```diff
@@ -37,18 +37,27 @@ func NewFiltersToMethods[MethodID comparable]() *FiltersToMethods[MethodID] {
 	}
 }
 
-// Registers the given method as having the given value for the given filter
-// kind.
-func (fm *FiltersToMethods[MethodID]) InsertNondirectionalFilter(filter FilterKind, value string, method MethodID) {
+// Registers the given filter and value in the map. These are also associated
+// with the method ID if it exists in the given set of allowed methods.
+func (fm *FiltersToMethods[MethodID]) InsertNondirectionalFilter(filter FilterKind, value string, method MethodID, allowedMethods sets.Set[MethodID]) {
```
thatplguy, line 42: This is fine, but I think it would be a little cleaner to change the type of `method` to `optionals.Optional[MethodID]` and let the caller decide when it should be Some or None.

thatplguy (approved):

spec_summary/filter_map.go
liujed: This PR mirrors the changes in #168. My initial approach there was indeed to pass option values, but I found myself duplicating the same code at most of the callers, so I figured I'd push it down into the implementation here.

liujed: I've opened akitasoftware/go-utils#11 to factor that common code into a method on `sets.Set`.

liujed: In 534259a.

liujed:
Re-requesting review for changes in 534259a.

thatplguy (approved):

PR #170, opened 2022-11-05
Refactor batcher.InMemory

liujed:
batcher.InMemory is now a generic wrapper around a Buffer that provides thread safety and manages flushing.

### DIFF
--- a/batcher/in_memory.go
+++ b/batcher/in_memory.go
@@ -8,19 +8,19 @@ import (
 )
 
 // A wrapper around a Buffer[Item] that manages thread-safety and flushing.
-type InMemory[Item any, Buf Buffer[Item]] struct {
-	buf Buf // protected by mu
+type InMemory[Item any] struct {
+	buf Buffer[Item] // protected by mu
 
 	signalClose                chan struct{} // closed when the batcher is closed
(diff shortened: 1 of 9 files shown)


batcher/in_memory.go
```diff
@@ -3,23 +3,25 @@ package batcher
...
+	"github.com/pkg/errors"
 )
 
-type InMemory struct {
-	processor  BatchProcessor
-	buf        []interface{} // protected by mu
-	maxBufSize int
+// A wrapper around a Buffer[Item] that manages thread-safety and flushing.
+type InMemory[Item any, Buf Buffer[Item]] struct {
+	buf Buf // protected by mu
```
mgritter, line 12: I don't understand how it benefits us to have InMemory take its buffer implementation as a type argument.  It's already an interface, so why is

```
  buf Buf
```

preferable to

```
  buf Buffer[Item]
```
?

mgritter (approved):
Looks OK; I see that the "real" buffer implementation is in the next PR.

I don't see why the buffer implementation type is a parameter, since the buffer object already an argument in the constructor and must conform to the interface type, so why not just treat it as the interface type?

batcher/in_memory.go
liujed: I was overly zealous when introducing those type parameters. Fixed in acacff3.

PR #171, opened 2022-11-08
Distinguish nullable from optional fields

thatplguy:
The Akita IR has an Optional node that is either Data (the field with this type is optional, i.e. may be present or absent) or None (the field with this type has only been seen with "null" as a value).

The Data node also has a "nullable" property, which indicates that a field with this type is capable of carrying "null" as a value.

Previously, we were treating the None node as optional, rather than nullable. That is, `meld(None, T) = Optional<T>`.  This PR fixes that, changing melding so that `meld(None, T) = Nullable<T>` where `Nullable<T>` means that `T.nullable = true`.

It also propagates `Data.nullable` during melding.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -202,9 +202,13 @@ type melder struct {
 //   - At most one variant in the OneOf is a list.
 //   - All other variants in the OneOf is a primitive.
 //
+// If the given src and dst have the following invariant on all OneOfs and
+// Optionals, then this is preserved.
+//
+//   - If any children are nullable, then the nullable bit is set on
+//     the parent, not the children.
+//
 // Assumes that dst.Meta == src.Meta.
-//
-//
 //
 // XXX: In some cases, this modifies src as well as dst :/
 func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -151,15 +151,27 @@ func mergeExampleValues(dst, src *pb.Data) {
 	dst.ExampleValues = examples
 }
 
-// Makes given Data optional if it isn't already.
+// Makes given Data optional if it isn't already.  If it is pb.None,
+// make it Optional<None>, which reflects that this field may be
+// absent (optional) and that it may also accept null values (None).
 func makeOptional(d *pb.Data) {
-	if !isOptional(d) {
+	isNone := d.GetOptional().GetNone() != nil
+	if !isOptional(d) || isNone {
+		// Maintain the invariant that the outermost Data object (the outer
+		// Optional, in this case) holds the nullable bit.
+		//
+		// d retains its nullable bit, even as its value is changed
+		// to be optional.
 		d.Value = &pb.Data_Optional{
 			Optional: &pb.Optional{
 				Value: &pb.Optional_Data{
 					Data: &pb.Data{Value: d.Value},
 				},
 			},
+		}
+
+		if isNone {
+			d.Nullable = true
 		}
 	}
 }
@@ -191,6 +203,8 @@ type melder struct {
 //   - All other variants in the OneOf is a primitive.
 //
 // Assumes that dst.Meta == src.Meta.
+//
+//
 //
 // XXX: In some cases, this modifies src as well as dst :/
 func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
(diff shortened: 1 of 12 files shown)


spec_util/testdata/meld/meld_nullable_1.pb.txt
```diff
@@ -0,0 +1,92 @@
...
+          key: "left_present_right_present"
+          value {
+            primitive {
+              int64_value {
+              }
+            }
+          }
+        }
+        fields {
+          key: "left_present_right_none"
```
liujed, line 21:
```suggestion
          key: "left_present_right_null"
```
Similarly for other places where "none" really means "present but null".

spec_util/meld_test.go
```diff
@@ -271,6 +272,26 @@ var tests = []testData{
...
+		"meld nullable and non-nullable versions of the same type",
+		[]string{
+			"testdata/meld/meld_nullable_1.pb.txt",
+			"testdata/meld/meld_nullable_2.pb.txt",
+		},
+		MeldOptions{},
+		"testdata/meld/meld_nullable_expected.pb.txt",
+	},
+	{
+		// Test meld(None, optional<T>) => nullable<optional<T>>
```
liujed, line 286: Should this be an `optional<nullable<T>>` instead?

liujed: Thinking on this more, in principle, `nullable<optional<T>>` doesn't make sense as a type: a `nullable<T>` isn't optional; it must be always present, even if null. Given this, I suppose we could let `nullable<optional<T>>` mean the same thing as `optional<nullable<T>>`, but having two representations of the same thing is a recipe for future headache.

This is just more evidence of a bad IR design. I don't know that there's much we can do here, beyond perhaps settling on a canonical representation of an optional nullable type.

thatplguy: Yea, it's not ideal.  The situation is worse for `oneof<T, ...>`, because there are combinatorially many equivalent types: `nullable<oneof<T, ...>>` and any combination of nullable children.  I chose to attach the nullable bit to the topmost IR node, which is why `nullable<optional<T>>`.  (Although I just realized that I need to lift the nullable bit when combining/creating oneofs.)

If I were to redesign things, I would make required/optional a property of structs rather than a first class type.  After all, what does `oneof<optional<T>, T2>` mean, or `list<optional<T>>`?  We could get rid of `optional` entirely and add a `null` type inhabited by the `null` value.  Then `optional<T>` would be syntactic sugar for `oneof<T, null>`, and we could flatten `oneof`s to get canonical types.

spec_util/testdata/meld/meld_nullable_1.pb.txt
thatplguy: I'd prefer not to change this, because "none" matches the syntax of the IR's None object.  Instead, I added more test fields with `not_present` (e.g. `left_present_right_not_present`) to indicate the field is not present, rather than present but null.

spec_util/meld.go
```diff
@@ -674,6 +697,11 @@ func (m *melder) meldOneOfVariant(dst *pb.OneOf, srcHash *string, srcVariant *pb
 		srcHash = nil
 	}
 
+	// Clear the nullable bit from srcVariant.  The nullable bit was
+	// copied to the dst Data object in meldData before calling
+	// meldOneOfVariant.
+	srcVariant.Nullable = false
```
liujed, line 707: The block above makes a copy of `srcVariant` before modifying it, and ensures that the block below rehashes the modified variant. Shouldn't we do something similar here?

spec_util/meld_test.go
thatplguy: 641cf703e351c4e5e6f72394f08f57c7b87ca55b updates the comment on `meldData` to explain the nullable invariant, and 260b2c8476adb4c31572fc8b7a156a9e1589f5eb ensures the invariant is maintained when melding optional and oneof nodes.

The latter also adds special handling for `Optional<None>`, which represents an optional field that, when present, has only been seen with a null value.

spec_util/testdata/meld/meld_nullable_expected.pb.txt
```diff
@@ -67,6 +94,19 @@ method {
...
+          key: "left_none_right_not_present"
+          value {
+            nullable: true
+            optional: {
+              data: {
+                optional: {
+                  none: {}
+                }
+              }
+            }
```
liujed, line 107: This makes sense...but wow. 🤢

spec_util/meld.go
thatplguy: Fixed in f90bb9b14868c883e1a2e71989fff523318e4939.

liujed (approved):

PR #181, opened 2022-12-02
Extend agent telemetry with agent CPU/memory usage

thatplguy:


api_schema/schema.go
```diff
@@ -524,6 +524,17 @@ type PostClientPacketCaptureStatsRequest struct {
...
 	ApidumpError     ApidumpErrorType `json:"apidump_error,omitempty"`
 	ApidumpErrorText string           `json:"apidump_error_text,omitempty"`
+
+	// Report CPU and memory usage, if available.
+	AgentUsage *AgentUsage `json:"agent_usage,omitempty"`
+}
+
+type AgentUsage struct {
+	// Akita agent CPU usage, as a percentage of total CPU availability.
+	RelativeCPU float64 `json:"relative_cpu"`
```
mgritter, line 534: This does not seem very helpful without context-- how many CPUs are in the container or VM or host?

I suppose the user might know the context already, but it seems confusing to have this be a relative number.

api_schema/schema.go
```diff
@@ -524,6 +524,17 @@ type PostClientPacketCaptureStatsRequest struct {
 	// Report on any errors encounted during apidump that should be shown to the user.
 	ApidumpError     ApidumpErrorType `json:"apidump_error,omitempty"`
 	ApidumpErrorText string           `json:"apidump_error_text,omitempty"`
+
+	// Report CPU and memory usage, if available.
+	AgentUsage *AgentUsage `json:"agent_usage,omitempty"`
```
liujed, line 529: `AgentResourceUsage` and `agent_resource_usage` would be more descriptive.

api_schema/schema.go
```diff
@@ -524,6 +524,17 @@ type PostClientPacketCaptureStatsRequest struct {
...
+}
+
+type AgentUsage struct {
+	// Akita agent CPU usage, as a percentage of total CPU availability.
+	RelativeCPU float64 `json:"relative_cpu"`
+
+	// Peak virtual memory size in bytes.
+	VMPeak uint64 `json:"vm_peak"`
```
liujed, line 537: This should be peak RSS.

liujed (approved):

api_schema/schema.go
thatplguy: Changed in 9a50cbd4827ff573e2d59bc69a8cfe7f73971278.

api_schema/schema.go
thatplguy: Fixed in 9a50cbd4827ff573e2d59bc69a8cfe7f73971278.

PR #183, opened 2023-01-07
Remove path patterns

liujed:
These are no longer used by the CLI as of akitasoftware/akita-cli#191.


mgritter:
These libraries are imported in superstar in several places.

mgritter:
Ah, I see, the linked PR switched to a superstar lib?

liujed:
Yes, this code is being replaced with code elsewhere.

versilis (approved):
Thanks for the explanation on why this is being removed. LGTM

PR #185, opened 2023-02-03
Fix melding of one-ofs with primitives

liujed:
When melding a primitive into a one-of, we need to meld against all compatible variants in the one-of, not just the first one we find. This allows the meld to collapse several variants into one, if the primitive has no formats.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -9,6 +9,7 @@ import (
 	"google.golang.org/protobuf/proto"
 
 	pb "github.com/akitasoftware/akita-ir/go/api_spec"
+	"github.com/akitasoftware/go-utils/optionals"
 	"github.com/akitasoftware/go-utils/sets"
 
 	"github.com/akitasoftware/akita-libs/spec_util/ir_hash"
@@ -228,9 +229,9 @@ func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
 	// Check if src is already a oneof. This can happen if src is the collapsed
 	// element from a list originally containing elements with conflicting types.
 	if srcOf, ok := src.Value.(*pb.Data_Oneof); ok {
-		if v, ok := dst.Value.(*pb.Data_Oneof); ok {
+		if _, ok := dst.Value.(*pb.Data_Oneof); ok {
 			// dst already encodes a conflict. Merge the conflicts.
-			return m.meldOneOf(v.Oneof, srcOf.Oneof)
+			return m.meldOneOf(dst, srcOf.Oneof)
 		}
 
 		// dst is not a oneof. Swap src and dst and re-use the logic below.
@@ -301,7 +302,7 @@ func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
 		}
 	case *pb.Data_Oneof:
 		hasConflict = true
-		return m.meldOneOfVariant(v.Oneof, nil, src)
+		return m.meldOneOfVariant(dst, optionals.None[string](), src)
 	default:
 		hasConflict = true
 		return m.recordConflict(dst, src)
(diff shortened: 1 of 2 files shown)


jombooth (approved):
LGTM (to the extent that I can review this)

spec_util/meld.go
```diff
@@ -721,37 +733,59 @@ func (m *melder) meldOneOfVariant(dst *pb.OneOf, srcHash *string, srcVariant *pb
...
+			srcHash = ir_hash.HashDataToString(srcVariant)
+		}
+
 	default:
 		return fmt.Errorf("unknown one-of variant type: %s", reflect.TypeOf(srcVariant.Value).Name())
 	}
 
+	if len(dstOneOf.Options) == 0 {
+		// No longer have a one-of. Replace dst's value with the singleton variant.
+		dst.Value = srcVariant.Value
```
mgritter, line 784: I am trying to think through all the cases that this new code might be triggered.  Is there a case where we need to re-hash dst where we would not have re-hashed previously?  (I don't think so, we should not have nested OneOf's.)

This new case seems to be exclusive to primitives, should we move it inside the case?  Or do you think this check is more broadly applicable?

spec_util/meld_test.go
```diff
@@ -614,6 +614,30 @@ func TestMeldData(t *testing.T) {
...
+					"1": wrapPrim(&pb.Primitive{
+						Value:      &pb.Primitive_Uint32Value{Uint32Value: &pb.Uint32{}},
+						FormatKind: "datetime",
+						Formats:    map[string]bool{"timestamp_seconds_since_epoch": true},
+					}),
+					"2": wrapPrim(&pb.Primitive{
+						Value:      &pb.Primitive_Uint64Value{Uint64Value: &pb.Uint64{}},
+						FormatKind: "unique_id",
+						Formats:    map[string]bool{"integer_id": true},
+					}),
```
mgritter, line 630: I would like to see the case where there is an incompatible type in the list as well, so that (string | datetime | unique_id) with (int) becomes (string | int).  i.e., where the new len(Options) == 0 branch is not taken, but some options have been deleted anyway.

mgritter (approved):
Looks OK.

spec_util/meld.go
liujed: I'll think about the first question, but to answer the second, I say we just leave this here. A singleton one-of isn't really a one-of and should be collapsed.

liujed: According to [this](https://github.com/akitasoftware/akita-libs/blob/c49ba8585c347d1927a2dd78fc9516d9807d7f19/spec_util/hash_keys.go#L9-L12), there are only three places in the IR where hashes are used: `Method.Args`, `Method.Responses`, and `OneOf.Options`. Any rehashing for the first two cases should already be handled as part of the contract on the meld procedure. And the third, as you say, shouldn't happen because we don't nest one-ofs.

spec_util/meld.go
```diff
@@ -721,37 +733,59 @@ func (m *melder) meldOneOfVariant(dst *pb.OneOf, srcHash *string, srcVariant *pb
...
+		// will add the resulting source variant to the destination.
+		needToRehashSrc := false
+		for dstHash, dstVariant := range dstOneOf.Options {
 			if _, dstIsPrim := dstVariant.Value.(*pb.Data_Primitive); dstIsPrim && haveCompatibleTypes(dstVariant.GetPrimitive(), srcVariant.GetPrimitive()) {
-				return m.meldAndRehashOption(dst, oldDstHash, dstVariant, srcVariant)
+				delete(dstOneOf.Options, dstHash)
+
+				// To avoid modifying *srcVariant, we meld srcVariant into dstVariant,
+				// and then set srcVariant = dstVariant.
+				if err := m.meldData(dstVariant, srcVariant); err != nil {
```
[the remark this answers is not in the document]
liujed: In 6e39308.

> I see that meldPrimitive has an "XXX" where it sometimes modifies src anyway :(

Yup. 😞 But at least this doesn't introduce any new code that violates the alleged contract!

spec_util/meld_test.go
liujed: In 4bcf3e4.

PR #186, opened 2023-02-04
Fix bug introduced in #185

liujed:
This addresses the following issue in `meldOneOf`. Suppose the source and destination both have multiple variants. During melding, a source variant can collapse the destination into a singleton. Previously, `meldOneOfVariant` would eagerly convert the destination into a Primitive. But if there are more source variants to meld, the next call to `meldOneOfVariant` would then yield an error because the destination is no longer a OneOf.

To address this, `meldOneOfVariant` no longer eagerly unwraps singleton OneOfs. Instead, this is now the caller's responsibility. The function `rewriteSingletonOneOf` is provided for this purpose.

### DIFF
--- a/spec_util/meld.go
+++ b/spec_util/meld.go
@@ -251,10 +251,12 @@ func (m *melder) meldData(dst, src *pb.Data) (retErr error) {
 			}
 			makeOptional(dst)
 			return nil
+
 		case *pb.Optional_None:
 			// If src is a none, drop the none and mark the dst value as nullable.
 			dst.Nullable = true
 			return nil
+
 		default:
 			return fmt.Errorf("unknown optional value type: %s", reflect.TypeOf(srcOpt.Optional.Value).Name())
 		}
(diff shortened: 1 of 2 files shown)


spec_util/meld.go
```diff
@@ -602,6 +616,18 @@ func stripOptional(data *pb.Data) *pb.Data {
 	return optional.GetData()
 }
 
+// Converts a singleton one-of into its underlying variant. Does nothing if the
+// given IR node is not a singleton one-of.
+func rewriteSingletonOneOf(data *pb.Data) {
```
jombooth, line 621: optional suggestion - a pattern I like for functions that might have no effect is to name them `maybeDoThing`, in this case `maybeRewriteSingletonOneOf`. This makes it super clear at call sites that the function may have no effect.

spec_util/meld_test.go
```diff
@@ -677,6 +677,53 @@ func TestMeldData(t *testing.T) {
 				PotentialConflict: true,
 			}),
 		},
+		{
+			// XXX The order in which a Go map is iterated changes from run to run.
+			// This test case only runs as intended when the variants in the
+			// right-hand side are merged in the order that they are written.
```
jombooth, line 683: Does this mean that the test will sometimes flake?

jombooth (approved):

mgritter (approved):

mgritter:
Hm, sorry I wasn't able to come up with this case in the previous review.  :(

I have been wanting to build a Alloy model of our IR to check things like this, but the opportunity is not likely to present itself any time soon.

PR #188, opened 2023-03-09
Add explicit types for hosts and endpoint paths

liujed:
Introduced an `Endpoint` type for representing REST endpoints, and its components: `HTTPMethod`, `Host`, and `PathTemplate`. Modified `EndpointGroup` so that it also uses these types.

### DIFF
--- a/http_rest_methods/methods.go
+++ b/http_rest_methods/methods.go
@@ -1,0 +1,65 @@
+package http_rest_methods
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/akitasoftware/go-utils/sets"
+	"github.com/pkg/errors"
+)
+
+type HTTPMethod string
+
+const (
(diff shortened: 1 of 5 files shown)


http_rest_methods/methods.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/akitasoftware/go-utils/sets"
+	"github.com/pkg/errors"
+)
+
+type HTTPMethod string
```
mgritter, line 11: I am really annoyed that we have to do this ourselves, the constants in net/http are just strings.

mgritter (approved):
Looks OK.

http_rest_methods/methods.go
liujed: Go: opinionated in all the wrong places! 🙃

PR #194, opened 2023-08-24
Add AKID for path parameter prefixes

versilis:
This PR adds a new AKID for path prameter prefixes. This will provide support for identifying path parameters edited by a user.

### DIFF
--- a/akid/akids.go
+++ b/akid/akids.go
@@ -8,31 +8,32 @@ import (
 )
 
 const (
-	APISpecTag            = "api"
-	APIKeyTag             = "apk"
-	APIMethodTag          = "mth"
-	ClientTag             = "cli"
-	ConnectionTag         = "cxn"
-	DataCategoryTag       = "dct"
-	GraphTag              = "gph"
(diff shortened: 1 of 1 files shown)


akid/akids.go
```diff
@@ -8,31 +8,32 @@ import (
...
+	DataCategoryTag        = "dct"
+	GraphTag               = "gph"
+	IdentityTag            = "idt"
+	InvalidTag             = "xxx"
+	LearnSessionTag        = "lrn"
+	MessageTag             = "msg"
+	MonitorTag             = "mtr"
+	OrganizationTag        = "org"
+	OutboundRequestTag     = "obr"
+	PathParameterPrefixTag = "pre"
```
liujed, line 25: (Non-blocking.) How about something more distinctive? There might be other kinds of prefixes. Granted, it's hard to compress "path parameter prefix" into just three letters.
```suggestion
	PathParameterPrefixTag = "ppx"
```

liujed (approved):

PR #195, opened 2023-09-01
Add marshall and unmarshall functions for path parameter prefix id

shreys7:


versilis (approved):

liujed (commented):
Looks good. We really ought to be generating this file.

PR #196, opened 2023-09-06
[OBS-291] Zero out primitive values instead of hashing them.

mgritter:
Deleted unused GeneralizeWitness function and its failing test.
Retained testdata/generalized_witnesses because it's in use by ir_hash/generated_types_test.go


mgritter:
I'll look at the failing test later; marking as draft until that's completed.

liujed (approved):

PR #199, opened 2023-09-28
[OBS-420] chore: pop first element from path template slice

mudit-postman:


mudit-postman:
Since this function is used in other places also so didn't change it

PR #201, opened 2023-11-02
Update akita-ir library

shreys7:
### Changes
This PR is built on top of https://github.com/akitasoftware/akita-ir/pull/10. 
- Here, we update the akita-ir version 
- In `akita-libs/spec_utils/ir_hash`, add the newly defined HTTPMethodMeta in `gen.go`
- Changes from running `make` in `akita-libs/spec_utils/ir_hash`

### TODO
Update akita-ir version to match the commit on master once [akita-ir#10](https://github.com/akitasoftware/akita-ir/pull/10) is merged

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.18
 
 require (
 	github.com/OneOfOne/xxhash v1.2.8
-	github.com/akitasoftware/akita-ir v0.0.0-20220630210013-8926783978fe
+	github.com/akitasoftware/akita-ir v0.0.0-20231102193917-8b96486d4377 // todo: update this to match the commit on master once akita-ir#10 is merged
 	github.com/akitasoftware/go-utils v0.0.0-20221207014235-6f4c9079488d
 	github.com/akitasoftware/objecthash-proto v0.0.0-20211020004800-9990a7ea5dc0
 	github.com/andreyvit/diff v0.0.0-20170406064948-c7f18ee00883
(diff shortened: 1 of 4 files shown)


mgritter (approved):
Looks correct, good to go once akita-ir changes land and the generator is re-run on that version.

akinet/net_traffic.go
```diff
@@ -374,6 +374,30 @@ func (tls *TLSHandshakeMetadata) ApplicationLatencyMeasurable() bool {
 	return *tls.SelectedProtocol == "http/1.1"
 }
 
+// Represents the content when a TCP connection is closed by the client.
+type ClientTimeoutMetadata struct {
```
mgritter, line 378: This looks OK, but it might be better to specifically say "Shutdown" rather than "Timeout" to be more specific.  (We don't _know_ that it timed out -- it might have crashed or been killed.  So timeout is an inference that might not be justified.)

mgritter (approved):

akinet/tcp_parser.go
```diff
@@ -14,6 +14,14 @@ const (
 	NeedMoreData
 )
 
+const (
+	CONNECTION_TYPE_HTTP_CLIENT   = "HTTP_CLIENT"
+	CONNECTION_TYPE_HTTP_SERVER   = "HTTP_SERVER"
+	CONNECTION_TYPE_TLS_CLIENT    = "TLS_CLIENT"
+	CONNECTION_TYPE_TLS_SERVER    = "TLS_SERVER"
+	CONNECTION_TYPE_HTTP2_PREFACE = "HTTP2_PREFACE"
+)
```
mgritter, line 23: Is it worth having these defined for each protocol type, or should we have only client and server?  (Or unknown?)

mgritter (approved):

PR #202, opened 2023-11-15
[POA-41] Integrate amplitude SDK and remove Mixpanel and SegmentIO SDKs

mudit-postman:
* Removed all the code related to segmentIO and Mixpanel
* Add support for Amplitude in analytics libs


mudit-postman:
I have removed the old code under the assumption that we will be migrating to Amplitude completely for all the Postman Agent releases
Let me know if we need to keep old integrations, maybe for Akita cli releases (if needed)

liujed:
My review will be focused on the addition of Amplitude. @mgritter, you have more context than I do on what is consuming the Mixpanel/Segment pieces, so I will need your help with the removal part of this PR.

analytics/client.go
```diff
@@ -21,175 +19,97 @@ type Client interface {
...
-			event.distinctID, event.name, &mixpanel.Event{
-				Properties: event.properties,
-				Timestamp:  &event.timestamp,
-			},
-		)
-		if mixpanelErr != nil {
-			err = multierror.Append(err, mixpanelErr)
-		}
-	}
+	c.amplitudeClient.Track(amplitude.Event{
```
mgritter, line 70: Hm, I can't believe they released an SDK that never returns any errors.

analytics/client.go
```diff
@@ -21,175 +19,97 @@ type Client interface {
...
-	for integrationName, enabled := range config.DefaultIntegrations {
-		integrations = integrations.Set(integrationName, enabled)
-	}
-
-	return integrations
+	c.amplitudeClient.Shutdown()
+	return nil
 }
 
 // Returns the logger to use for the segment client if logging is enabled. Otherwise, returns nil.
```
mgritter, line 89:
```suggestion
// Returns the logger to use for the Amplitude client if logging is enabled.
```

mgritter (approved):
Looks good.

mgritter:
I thought about this a bit more.

The removal of Segment makes it all-or-nothing, which is tricky if we want to release a CLI while this is half done.  Superstar and the CLI must of necessity share the same akita-libs module.

It would be somewhat better, on that axis, if the Amplitude constructor was its own function that returned an instance of the same interface (which I don't think changed?), so we could run the CLI with Amplitude and the superstar services with the older analytics package.

mudit-postman:
> Superstar and the CLI must of necessity share the same akita-libs module.
Is there any issue if they are on a different version? Also I will be shifting from Segment to Amplitude in superstar service also, so we can release both of them with same version of akita-libs

mgritter:
Because superstar is a monorepo it shared a single go.mod for all services.  That same go.mod includes akita-cli, because we build the agent out of the superstar tree too.  Thus, if akita-cli needs a new akita-libs to build, all services are necessarily upgraded to the same akita-libs version.  (This is a problem with Go modules, but also with the decision that all services were part of a single module.)

We could build akita-cli in a branch if we really had to, I think, but this might mess up the automation.

mudit-postman:
Closing since, merged #203 PR instead of this.

PR #203, opened 2023-11-22
feat: Integrate amplitude SDK along other analytics SDKs

mudit-postman:
Base on the discussion on #202, this PR now integrates Amplitude SDK along with the other SKDs to support backward compatability


analytics/client.go
```diff
@@ -29,33 +30,15 @@ type clientImpl struct {
 
 	// TODO: Remove Mixpanel once we've confirmed that Segment is working.
 	mixpanelClient mixpanel.Mixpanel
+
+	// The internal client used to send events to Amplitude.
+	amplitudeClient amplitude.Client
+	// App info from which event is sent
+	amplitudeAppInfo amplitude.EventOptions
```
liujed, line 37: This comment is hard to understand. I have a suggestion for improvement below. However, `amplitudeClient` really belong together `amplitudeAppInfo`, so we'd be better served by having an abstraction that wraps them both and injects `amplitudeAppInfo` into all reported tracking events. **Non-blocking.**

```suggestion
	// This is included in all tracking events reported to Amplitude.
	amplitudeAppInfo amplitude.EventOptions
```

analytics/client.go
```diff
@@ -149,6 +167,69 @@ func newMixpanelClient(config Config) (mixpanel.Mixpanel, error) {
 	return mixpanel.New(config.MixpanelToken, mixpanelURL), nil
 }
 
+func newAmplitudeClient(config Config) (amplitude.Client, amplitude.EventOptions, error) {
+	if !config.IsAmplitudeEnabled || config.AmplitudeConfig == (AmplitudeConfig{}) {
```
liujed, line 171: Why provide two different ways of disabling the Amplitude client? If the configuration has explicitly specified that Amplitude is enabled, shouldn't the absence of a configuration be an error? **Blocking.**

analytics/client.go
```diff
@@ -205,3 +295,42 @@ func (d disabledLogger) Logf(format string, args ...any) {
 func (d disabledLogger) Errorf(format string, args ...interface{}) {
 	// Do nothing.
 }
+
+// A custom segment logger that logs to glog.
+type amplitudeLogger struct{}
```
liujed, line 300: This file is starting to feel a bit cluttered. We should separate the code for Mixpanel, Segment, and Amplitude out into separate files. **Non-blocking.**

liujed (approved):
Looks good. A few minor comments. Please address at least the one blocking comment before merging.

analytics/client.go
mudit-postman: I have not done it cause we might be removing the segment and Mixpanel SDKs when we are sure that Amplitude is working fine and all the metrics are passed to Amplitude.

PR #205, opened 2024-01-02
[POA-484] Add PostmanUser struct in UserResponse

mudit-postman:
For sending postman's userID as Amplitude telemetry's userID we need to pass it to the Akita CLI.
So added `postmanUser` to the `UserResponse` model, used in Akita-cli


mgritter (approved):
Is this still relevant?  Approved anyway.  :)

PR #213, opened 2024-07-02
POA-1341 Update IR

liujed:
Excludes the path-parameter algorithm annotation from Methods' hashes.

### DIFF
--- a/spec_util/ir_hash/gen/gen.go
+++ b/spec_util/ir_hash/gen/gen.go
@@ -29,6 +29,7 @@ func main() {
 	gf.SetPackageName("ir_hash")
 	gf.AddImports()
 
+	gf.AddIgnoredField("HTTPMethodMeta", "PathParamAlg")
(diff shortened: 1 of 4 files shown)


spec_util/ir_hash/gen/gen.go
```diff
@@ -29,6 +29,7 @@ func main() {
 	gf.SetPackageName("ir_hash")
 	gf.AddImports()
 
+	gf.AddIgnoredField("HTTPMethodMeta", "PathParamAlg")
```
mgritter, line 32: Hm, this is an argument I'd forgotten about not putting things in the IR -- we want to be able to hash it to check for equality.  (Which is one reason I regret putting ProcessingLatency here.)

I think not including it in the hash is correct, though. Two methods with identical content but different PathParamAlg is correct.

mgritter (approved):
Looks good.

PR #214, opened 2024-08-02
[POA-1519] Allow HTTP response without reason phrase

mgritter:
Technically, the RFC requires a space even if the reason phrase is empty, but at least one service has been observed to violate this. Add unit tests for both correct and incorrect blank reason phrase, and a one reproducing the original error to verify that Go's http parser is fine with this.

### DIFF
--- a/akinet/http/parser_factory.go
+++ b/akinet/http/parser_factory.go
@@ -155,7 +155,11 @@ func hasValidHTTPResponseStatusLine(input memview.MemView) akinet.AcceptDecision
 
 	// A space separates the HTTP version from status code.
 	// The format is SP Status-Code SP Reason-Phrase CR LF
-	if input.GetByte(0) != ' ' || input.GetByte(4) != ' ' {
+	// Some servers do not include the second 'SP" though and skip straight to 'CR'.
+	if input.GetByte(0) != ' ' {
+		return akinet.Reject
+	}
+	if input.GetByte(4) != ' ' && input.GetByte(4) != '\r' {
 		return akinet.Reject
 	}
 
--- a/akinet/http/parser_factory.go
+++ b/akinet/http/parser_factory.go
@@ -149,7 +149,7 @@ func hasValidHTTPRequestLine(input memview.MemView) akinet.AcceptDecision {
 // RFC 2616 Section 6.1. The input should start right after the HTTP version.
 func hasValidHTTPResponseStatusLine(input memview.MemView) akinet.AcceptDecision {
 	if input.Len() < 5 {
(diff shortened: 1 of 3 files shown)


shreys7 (approved):

shreys7:
> Technically, the RFC requires a space even if the reason phrase is empty, but at least one service has been observed to violate this.
Can you add where we observed this and how you found that this was the case? Empty reason phrase sounds skeptical, since I don't think we would have added any special handling in the Application layer for the service to do this.

akinet/http/parser_factory_test.go
```diff
@@ -226,6 +226,18 @@ func TestHTTPResponseParserFactoryAccepts(t *testing.T) {
 			expectedDecision: akinet.Accept,
 			expectedDF:       int64(len("OK")),
 		},
+		// This is technically illegal but we observed some Postman services sending it.
+		{
+			name:             "no reason message",
+			input:            "HTTP/1.1 200\r\n",
+			expectedDecision: akinet.Accept,
+		},
+		// This is technically legal, as the reason string can be zero bytes long.
```
shreys7, line 235: Or we can just right `This is technically legal as reason string is optional` for better readability.

mgritter:
> > Technically, the RFC requires a space even if the reason phrase is empty, but at least one service has been observed to violate this.
> 
> Can you add where we observed this and how you found that this was the case? Empty reason phrase sounds skeptical, since I don't think we would have added any special handling in the Application layer for the service to do this.

The TCPdumps demonstrating the problem are in https://postmanlabs.atlassian.net/browse/POA-1519; I didn't want to be more specific about Postman internal services in a public repository.

akinet/http/parser_factory_test.go
mgritter: I think it is less clear what "optional" means than the specific thing this test is demonstrating

akinet/http/parser_factory.go
```diff
@@ -155,7 +155,11 @@ func hasValidHTTPResponseStatusLine(input memview.MemView) akinet.AcceptDecision
 
 	// A space separates the HTTP version from status code.
 	// The format is SP Status-Code SP Reason-Phrase CR LF
-	if input.GetByte(0) != ' ' || input.GetByte(4) != ' ' {
+	// Some servers do not include the second 'SP" though and skip straight to 'CR'.
```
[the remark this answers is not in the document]
mgritter: Updated in https://github.com/postmanlabs/observability-shared-libs/pull/214/commits/c68fa8cb26fff4b7352fb137af77321416921f84

PR #215, opened 2024-08-15
POA-1463 Remove dependency on martian

liujed:
The `martian` package is no longer maintained, and was only used for its HAR data structures. HAR is no longer relevant to us.


tags/tags.go
```diff
@@ -3,28 +3,27 @@ package tags
 import (
 	"strings"
 
-	"github.com/google/martian/v3/tags"
 	"github.com/pkg/errors"
 )
 
-type Key = tags.Key
+type Key = string
```
mgritter, line 9: The martian code (added by you) was

```
type Key string
```

shouldn't this be a type definition, not an alias, so that we retain a non-string type?

mgritter (approved):
This is fine but I am surprised you didn't stick with 

```
Set( key Key, values []string )
```

etc?

tags/tags.go
liujed: Yeah, I missed that it was a typedef, not an alias. In 9abc4fe.

PR #219, opened 2024-11-12
POA-2294 Include names of multi-part bodies in visitor context

liujed:


visitors/http_rest/spec_visitor.go
```diff
@@ -514,69 +515,109 @@ func extendContext(cin Context, node interface{}) {
...
+			isMultipartBodyMember := false
+			multipartBodyMemberNameOpt := optionals.None[string]()
+			if meta.GetBody() != nil {
+				outerData, _ := ctx.GetInnermostNode(reflect.TypeOf((*pb.Data)(nil)))
+				if outerData != nil {
+					isMultipartBodyMember = outerData.(*pb.Data).GetMeta().GetHttp().GetMultipart() != nil
+					if isMultipartBodyMember {
+						// Get the name of the multipart body.
+						astPath := ctx.GetPath()
+						if !astPath.IsEmpty() {
```
mgritter, line 539: If it _is_ empty, what does that mean?

visitors/http_rest/spec_visitor.go
```diff
@@ -514,69 +515,109 @@ func extendContext(cin Context, node interface{}) {
...
-				ctx.appendRestPath(responseCode)
-				ctx.setResponseCode(responseCode)
 			}
 
 			// Figure out the name and kind of parameter being visited. If visiting a
 			// body, also figure out its content type.
-			var name string
-			named := true
+			restPathNames := []string{}
+			fieldPathName := optionals.None[string]()
```
mgritter, line 574: I seem to recall the context path was an allocation hot-spot, so if this array allocation could be avoided that would be desirable, but I don't see a good way to do it other than move RestPath manipulation into each of the cases.

visitors/http_rest/spec_visitor.go
```diff
@@ -514,69 +515,109 @@ func extendContext(cin Context, node interface{}) {
...
+		if meta := node.GetMeta().GetHttp(); meta != nil {
+			// Multipart bodies have two layers of metadata: an outer layer associated
+			// with the Data node representing the multiple parts, and an inner layer
+			// at the Data nodes representing each individual part. Determine whether
+			// we are at the inner layer, and if so, try to get the name of the
+			// current body part.
+			//
+			// ugh.
+			isMultipartBodyMember := false
+			multipartBodyMemberNameOpt := optionals.None[string]()
```
mgritter, line 531: I was trying to think of alternatives here.

It seems like we want to know something about the enclosing Data while visiting this Data.  Instead of walking back up to find that Data object, would it make sense to signal "isMultipart" as a new field in the context?  (I think this is not a clear solution because we might want to clear it out at the next level of Data?  Or manipulate it only for Bodys?)

Or, we could keep track of the enclosing GetMeta().GetHttp(), because that is what we are trying to get?

mgritter (approved):

visitors/http_rest/spec_visitor.go
liujed: Possibly that the visitor was started at a point lower in the IR that doesn't include this information.

PR #230, opened 2025-04-29
Allow PacketCounts to be reset

cesher:
I ran into a couple of roadblocks with previous plan for calculating per agent throughput and total throughput with just the cumulative count. I have settled on the following plan:

- Introduce a new set of counters, `ObservationWindow` though open to alternative names, that reset each time we send a `PacketCountSummary`.
- Use the counts in the observation window to build a time series table.
- Query the time series table over a time period to determine the number of agents that reported rate limited request and sum the number of requests processed in that time period.

### DIFF
--- a/client_telemetry/pcap_stats.go
+++ b/client_telemetry/pcap_stats.go
@@ -61,7 +61,7 @@ func (c *PacketCounts) CopyAndReset() *PacketCounts {
 // Reflects the version of the JSON encoding.  Increase the minor version
 // number for backwards-compatible changes and the major number for non-
 // backwards compatible changes.
-const Version = "v0.3"
+const Version = "v0.4"
 
 type PacketCountSummary struct {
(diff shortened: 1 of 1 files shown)


client_telemetry/pcap_stats.go
```diff
@@ -42,17 +42,34 @@ func (c *PacketCounts) Copy() *PacketCounts {
...
+	return copy
+}
+
 // Reflects the version of the JSON encoding.  Increase the minor version
 // number for backwards-compatible changes and the major number for non-
 // backwards compatible changes.
 const Version = "v0.3"
 
 type PacketCountSummary struct {
-	Version        string                   `json:"version"`
```
mgritter, line 51: Consider bumping the minor version number to signal the presence of ObservationWindow?

mgritter (approved):

